PR #21, opened 2015-03-28
OkHttp: Fix NetworkOnMainThreadException while cancellation

yongjhih:
See square/okhttp#869 also.

### DIFF
--- a/imagepipeline-backends/imagepipeline-okhttp/src/main/java/com/facebook/imagepipeline/backends/okhttp/OkHttpNetworkFetcher.java
+++ b/imagepipeline-backends/imagepipeline-okhttp/src/main/java/com/facebook/imagepipeline/backends/okhttp/OkHttpNetworkFetcher.java
@@ -10,6 +10,7 @@
 package com.facebook.imagepipeline.backends.okhttp;
 
 import android.net.Uri;
+import android.os.Looper;
 
 import com.facebook.common.logging.FLog;
 import com.facebook.common.references.CloseableReference;
@@ -27,6 +28,7 @@ import com.squareup.okhttp.Response;
 import com.squareup.okhttp.ResponseBody;
 
 import java.io.IOException;
+import java.util.concurrent.Executor;
 
 /**
  * Network fetcher that uses OkHttp as a backend.
@@ -37,11 +39,14 @@ public class OkHttpNetworkFetcher extends BaseNetworkFetcher<FetchState> {
 
   private final OkHttpClient mOkHttpClient;
 
+  private final Executor mCancellationExecutor;
+
   /**
    * @param okHttpClient client to use
    */
   public OkHttpNetworkFetcher(OkHttpClient okHttpClient) {
     mOkHttpClient = okHttpClient;
+    mCancellationExecutor = okHttpClient.getDispatcher().getExecutorService();
   }
 
   @Override
@@ -65,7 +70,15 @@ public class OkHttpNetworkFetcher extends BaseNetworkFetcher<FetchState> {
         new BaseProducerContextCallbacks() {
           @Override
           public void onCancellationRequested() {
-            call.cancel();
+            if (Looper.myLooper() != Looper.getMainLooper()) {
+              call.cancel();
+            } else {
+              mCancellationExecutor.execute(new Runnable() {
+                @Override public void run() {
+                  call.cancel();
+                }
+              });
+            }
           }
         });
 


imagepipeline-backends/imagepipeline-okhttp/src/main/java/com/facebook/imagepipeline/backends/okhttp/OkHttpNetworkFetchProducer.java
```diff
@@ -83,7 +85,17 @@ protected void fetchImage(final NfpRequestState requestState) {
           new BaseProducerContextCallbacks() {
             @Override
             public void onCancellationRequested() {
-              call.cancel();
+              // Avoid NetowkrOnMainThreadException
+              if (Looper.myLooper() != Looper.getMainLooper()) {
+                call.cancel();
+              } else {
+                new AsyncTask<Call, Void, Void>() {
```
jasta: `AsyncTask` is overkill here.  Specifically, `AsyncTask` is needed only when you require a response in some fashion from the background job on a Handler-based thread (typically the main thread).  In this case, you have fire-and-forget semantics, which a simple Thread or Executor will do much better.

I'm not as familiar with Fresco's code, but I suspect there is a standard internal Executor (or something just like it) that can be used for this purpose as well.

imagepipeline-backends/imagepipeline-okhttp/src/main/java/com/facebook/imagepipeline/backends/okhttp/OkHttpNetworkFetchProducer.java
```diff
@@ -83,7 +85,17 @@ protected void fetchImage(final NfpRequestState requestState) {
           new BaseProducerContextCallbacks() {
             @Override
             public void onCancellationRequested() {
-              call.cancel();
+              // Avoid NetowkrOnMainThreadException
```
jasta: Typo here and in your commit message.

yongjhih: Thanks for correction

tyronen:
Hi Andrew, thanks for doing this.

I see here though that, if the Executor is not supplied, you're creating a whole new thread pool just for cancellations. 

OkHttpClient itself doesn't seem to expose its thread pool, nor offer an async cancel method which would be a proper solution. 

A workaround might be to use [ExecutorSupplier.forBackground](http://frescolib.org/javadoc/reference/com/facebook/imagepipeline/core/ExecutorSupplier.html#forBackground\(\)) method.

What you can do is make an instance of this an optional argument to OkHttpImagePipelineConfigFactory.newBuilder. If it isn't supplied, use DefaultExecutorSupplier. Then pass it to OkHttpNetworkFetchProducer, and before returning call ImagePipelineConfig.Builder.setExecutorSupplier.

jasta:
I believe okhttp does expose its executor via
client.getDispatcher().getExecutorService().  That's probably your easiest
option to fix this.

yongjhih:
Thanks @tyronen and @jasta. Indeed, I thought options of ExecutorSupplier.forBackground() and client.getDispatcher().getExecutorService() but I didn't finish it.

First, it's complicated to pass executorSupplier to okHttpNetworkFetchProducer.
OkHttpImagePipelineConfigFactory provides an imagePipelineConfigBuilder that already setup OkHttpNetworkFetchProducer. How to pass executorSupplier to okHttpNetworkFetchProducer when ImagePipelineConfigBuilder.setExecutorSupplier(executorSupplier)? Inherit ImagePipelineConfigBuilder?

I'm not as familiar with okhttpclient.getDispatcher().getExecutorService() so I have not tried yet.

tyronen:
client.getDispatcher is a better solution anyway. The network library's threads should be used for network operations.

yongjhih:
Thanks @tyronen, I tested it, it works. So I updated PR.

tyronen:
Hi sorry for the delay in getting back to you, it was Easter holiday. We have some other changes in head that you'll need to rebase off of.

yongjhih:
@tyronen, It's ok. I rebased.

tyronen:
LGTM

imagepipeline-backends/imagepipeline-okhttp/src/main/java/com/facebook/imagepipeline/backends/okhttp/OkHttpNetworkFetcher.java
```diff
@@ -37,11 +39,14 @@
 
   private final OkHttpClient mOkHttpClient;
 
+  private Executor mCancellationExecutor;
```
DHuckaby: This field should be final

yongjhih: It's rebasing mistake, Thanks.

PR #53, opened 2015-03-31
[Supported URIs] makes local resource uri creation easier and "safer"

tbarthel-fr:
Don't know if it's a dummy request or not... maybe it could be useful for other devs.

As we didn't found any method to ensure a valid and supported URI for local resources, we tried to add one.

Should .gif be added to supported local resource extensions ?


fbcore/src/main/java/com/facebook/common/util/UriUtil.java
```diff
@@ -42,6 +45,12 @@
   public static final String LOCAL_RESOURCE_SCHEME = "res";
 
   /**
+   * Local resource supported extensions
+   */
+  private static final String LOCAL_RESOURCE_EXT_PNG = ".png";
+  private static final String LOCAL_RESOURCE_EXT_JPG = ".jpg";
```
plamenko: Not sure that we should rely on the extension.

tbarthel-fr:
Hi,

First of all , thanks a lot for considering this PR despite local images might not be a priority.

As you suggested, the imaging-specific code from fbcore has been removed and a new method (newBuilderWithResourceId) has been added.

Extension is definitively not reliable to determine the format, the first idea was to try to encapsulate the behaviour explained in the "Supported URIs" documentation : "Important thing to note is that this is not an image. If you want to display an XML drawable as the main image, then set it as a placeholder and use the null uri." But placeholders can't be specified at the ImageRequest level. Would it make sense to look at how to do that ?

imagepipeline/src/main/java/com/facebook/imagepipeline/request/ImageRequestBuilder.java
```diff
@@ -27,6 +30,12 @@
  */
 public class ImageRequestBuilder {
 
+  /**
+   * Local resource supported extensions
+   */
+  private static final String LOCAL_RESOURCE_EXT_PNG = ".png";
+  private static final String LOCAL_RESOURCE_EXT_JPG = ".jpg";
```
tyronen: Android supports gifs too, although these are discouraged. You could simply put these into a HashSet instead.

imagepipeline/src/main/java/com/facebook/imagepipeline/request/ImageRequestBuilder.java
```diff
@@ -47,6 +56,34 @@ public static ImageRequestBuilder newBuilderWithSource(Uri uri) {
...
+   * <p/>
+   * See also : {@link com.facebook.common.util.UriUtil#isLocalResourceUri(android.net.Uri)}
+   *
+   * @param resources application resources.
+   * @return a new request builder instance.
+   */
+    TypedValue value = new TypedValue();
+    resources.getValue(resId, value, true);
+    String resourceName = value.string.toString();
+    String resourceExt = resourceName.substring(resourceName.lastIndexOf('.'));
```
tyronen: Make this lowercase - on some filesystems casing is an issue.

imagepipeline/src/main/java/com/facebook/imagepipeline/request/ImageRequestBuilder.java
```diff
@@ -47,6 +56,34 @@ public static ImageRequestBuilder newBuilderWithSource(Uri uri) {
...
+
+    if (!resourceExt.equals(LOCAL_RESOURCE_EXT_PNG) && !resourceExt.equals(LOCAL_RESOURCE_EXT_JPG)) {
+      throw new IllegalArgumentException("Only image resources can be used with the image"
```
tyronen: `"Only image resources can be used with the image pipeline. " + resourceExt + " files are not supported."`

gradle.properties
```diff
@@ -42,6 +42,7 @@ POWERMOCK_MODULE_JUNIT4_RULE_VERSION=1.6.1
 ROBOLECTRIC_VERSION=2.4
 SUPPORT_V4_VERSION=21.0.3
 VOLLEY_VERSION=1.0.6
+ANDROID_ANNOTATIONS_VERSION=21.0.3
```
tyronen: Also please preserve the alphabetical order in the file.

tyronen:
Guava in its current form is not suitable for mobile, though if a mobile version were prepared it would indeed be very elegant. We do have a very small fork of it in fbcore's common/internal package, but it doesn't have ImmutableSet. To do that properly means using its immutability to gain efficiency, a nontrivial problem beyond the scope of the Fresco project.

For now I think you're good using a regular Java Set, as this isn't being passed around. You'll also need to rebase off head.

tbarthel-fr:
Thanks for your precious feedback!

"feature-branch" should be rebased and small changes should be applied.

PR #54, opened 2015-03-31
Using "provided" rather than "compile" for annnotations libraries.

mbrubin56:
Although these are necessary for the compilation of the product, they are not needed at runtime and need not be/should not be shipped with the library. The change is semantically more correct, produces a better result, and decreases the chances of causing a dex issue with multiply defined symbols.


tyronen:
This has been merged.

PR #70, opened 2015-04-01
Update ProducerSequenceFactory.java

karthikrg:
Need to handle the case where the uri string is shorter than 30 chars.

### DIFF
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ProducerSequenceFactory.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ProducerSequenceFactory.java
@@ -192,8 +192,12 @@ public class ProducerSequenceFactory {
     } else if (UriUtil.isLocalResourceUri(uri)) {
       return getLocalResourceFetchSequence();
     } else {
+      String uriString = uri.toString();
+      if (!TextUtils.isEmpty(uriString) && uriString.length() > 30) {
+        uriString = uriStirng.substring(0, 30);
+      }
       throw new RuntimeException(
-          "Unsupported image type! Uri is: " + uri.toString().substring(0, 30));
+          "Unsupported image type! Uri is: " + uriString);
     }
   }
 
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ProducerSequenceFactory.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ProducerSequenceFactory.java
@@ -193,8 +193,8 @@ public class ProducerSequenceFactory {
       return getLocalResourceFetchSequence();
     } else {
       String uriString = uri.toString();
-      if (!TextUtils.isEmpty(uriString) && uriString.length() > 30) {
-        uriString = uriStirng.substring(0, 30);
+      if (!TextUtils.isEmpty(uriString)) {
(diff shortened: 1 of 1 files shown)


imagepipeline/src/main/java/com/facebook/imagepipeline/core/ProducerSequenceFactory.java
```diff
@@ -192,8 +192,12 @@ private static void validateEncodedImageRequest(ImageRequest imageRequest) {
     } else if (UriUtil.isLocalResourceUri(uri)) {
       return getLocalResourceFetchSequence();
     } else {
+      String uriString = uri.toString();
```
tyronen: Thanks for sending this!

I think you can do this more simply with

```java
uriString.substring(0, Math.min(30, uriString.length));
```

karthikrg: Fixed. Thanks.

imagepipeline/src/main/java/com/facebook/imagepipeline/core/ProducerSequenceFactory.java
```diff
@@ -192,8 +192,12 @@ private static void validateEncodedImageRequest(ImageRequest imageRequest) {
...
+      if (!TextUtils.isEmpty(uriString)) {
+        uriString.substring(0, Math.min(30, uriString.length));
```
plamenko: Hmm, I wonder how did we end up with 30 in the first place... Uri's are usually long, and the extension is usually at the end of them.

tyronen:
Also, have you signed the CLA?

karthikrg:
Yes I have signed the CLA. I don't have write access to commit though.

tyronen:
This page gives instructions on how to "squash" your two commits into one: http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit

imagepipeline/src/main/java/com/facebook/imagepipeline/core/ProducerSequenceFactory.java
```diff
@@ -192,8 +192,12 @@ private static void validateEncodedImageRequest(ImageRequest imageRequest) {
...
+      }
       throw new RuntimeException(
-          "Unsupported image type! Uri is: " + uri.toString().substring(0, 30));
+          "Unsupported image type! Uri is: " + uriString);
```
plamenko: This is really "uri scheme", not "image type" so we should change the message too.

imagepipeline/src/main/java/com/facebook/imagepipeline/core/ProducerSequenceFactory.java
plamenko: Okay, because the error message below was not worded correctly, I misunderstood the code. Given that we want to output the uri scheme, first 30 characters is fine.

Instead of using `isEmpty` and then `Math.min`, it is better to do `if (uri != null && uri.length > 30)`.

tyronen:
Hi, thanks again for finding this bug. We ended going with a slightly different version in [this](https://github.com/facebook/fresco/commit/bdc3f167b6be879713a4074ab983eb7b9faed687) commit.

For some reason your CLA signature is not getting picked up by our bot - I'll see if I can dig into that.

PR #87, opened 2015-04-03
Fix small grammatical mistake in readme

tpalsulich:
This is a trivial change -- the list of features in the readme should end with "and much more," instead of, "and many more."


tyronen:
Accepted and exported.

PR #112, opened 2015-04-08
[Build] fix windows build

tbarthel-fr:
ndk-build.cmd found in $PATH , but internal batch are apparently called with relative path. 

Using ndk-folder absolute path from gradle plugin seems to fix it.

:imagepipeline:copyGiflib
:imagepipeline:ndk_build_gifimage'"C:\Users\tbarthel\Documents\GitHub\fresco\imagepipeline\find-win-host.cmd"' not recognized as an internal or external command, operable program or batch file.'
 FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':imagepipeline:ndk_build_gifimage'.
> Process 'command 'ndk-build.cmd'' finished with non-zero exit value 1

Could we only rely on the getNdkFolder or would it break "gradlew build" on other OS ?

### DIFF
--- a/imagepipeline/build.gradle
+++ b/imagepipeline/build.gradle
@@ -6,6 +6,8 @@ version = VERSION_NAME
 
 apply plugin: 'robolectric'
 apply plugin: 'de.undercouch.download'
+
+import com.android.build.gradle.LibraryPlugin
 import de.undercouch.gradle.tasks.download.Download
 import com.palominolabs.gradle.task.git.clone.GitCloneTask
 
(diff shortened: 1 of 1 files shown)


michalgr:
Hi ! Have you tried setting up 'ndk.path' in C:\Users\tbarthel\.gradle\gradle.properties ? Does that work for you ?

tbarthel-fr:
Hi!

C:\Users\tbarthel.gradle\gradle.properties didn't exist.

I create one with the right ndk.path and it build correctly (:

I presume that it's more reliable than using ndk.dir from the project local.properties, isn't it ?

Anyway, thanks a lot (:

PR #121, opened 2015-04-09
64-bit ARM and x86_64 architecture support (NB: not yet fully tested!)

orospakr:
Needed for the recently released Galaxy S6, among a few other models.

The crash you'd see would be something like:

    E/AndroidRuntime(5325): java.lang.UnsatisfiedLinkError: dalvik.system.PathClassLoader[DexPathList[[zip file "/data/app/com.joistapp.android.joist.alpha-1/base.apk"],nativeLibraryDirectories=[/data/app/com.joistapp.android.joist.alpha-1/lib/arm64, /vendor/lib64, /system/lib64]]] couldn't find "libgnustl_shared.so"

* Added the x86_64 and arm64-v8a architectures.
* Fixed pointer-length disagreement cast build failures in gifimage and webp.

I followed the guidance given at http://developer.android.com/training/articles/perf-jni.html.

Unfortunately, I lack a 64-bit ARM device, and the x86_64 emulator is not cooperating with me at the moment, so the best I've done is test for cursory regression in the context of my own app on a 32-bit device (which, I'll note, does not use either GIF or WebP).


orospakr:
My CTO is looking at the CLA.

orospakr:
Yep, just confirmed that this gets the core functionality of Fresco working on 64-bit phones.  WebP and GIF remain untested, however.

orospakr:
CLA is signed.

orospakr:
Will address the code review shortly!

plamenko:
@orospakr thanks!

xialin:
any progress on this?

michalgr:
We are merging https://github.com/facebook/fresco/pull/176

PR #122, opened 2015-04-09
Updating README to use version 0.2.0 to match docs

robert-wallis:
The docs here recommend 0.2.0+
http://frescolib.org/docs/download-fresco.html

However the README recommends 0.1.0+


tyronen:
LGTM

mdp:
:+1:
Just got bit by a 0.1.0 bug :) Would be nice to get this bumped for future readers. Great library by the way!

PR #212, opened 2015-04-28
a tiny improvement on execuating performance

LyndonChin:
This if statement `if (context instanceof ListenableActivity)` will be executed only once on this improvement.

### DIFF
--- a/fbcore/src/main/java/com/facebook/common/activitylistener/ActivityListenerManager.java
+++ b/fbcore/src/main/java/com/facebook/common/activitylistener/ActivityListenerManager.java
@@ -32,13 +32,12 @@ public class ActivityListenerManager {
   public static void register(
       ActivityListener activityListener,
       Context context) {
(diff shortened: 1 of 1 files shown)


fbcore/src/main/java/com/facebook/common/activitylistener/ActivityListenerManager.java
```diff
@@ -32,13 +32,12 @@
...
       ActivityListener activityListener,
       Context context) {
-    if (!(context instanceof ListenableActivity) && context instanceof ContextWrapper) {
-      context = ((ContextWrapper) context).getBaseContext();
-    }
     if (context instanceof ListenableActivity) {
       ListenableActivity listenableActivity = (ListenableActivity) context;
       Listener listener = new Listener(activityListener);
       listenableActivity.addActivityListener(listener);
+    } else if (context instanceof ContextWrapper) {
```
tyronen: This is not the same as the original behavior. If a ListenableActivity is wrapped in a ContextWrapper, the listener won't get added in this case.

PR #375, opened 2015-06-15
Add missing language for documentation

kunny:
Added missing available language(Korean) for the documentation in README.


tyronen:
Oh sorry. I just pushed out a similar fix. Thanks for noticing this!

PR #384, opened 2015-06-17
DefaultZoomableController: Add max scale factor

kunny:
Added max scale factor and setter for each min/max scale factors to DefaultZoomableController.


tyronen:
We've already added this and it will be pushed soon.

PR #436, opened 2015-07-08
fix compile error for NDK

Juude:


michalgr:
What compile error do you see? I wonder why didn't we spot it yet.

Juude:
first  I  did not compiled ok util set version to 4.9
but now I cannot reproduce the error.
I will research the process when found the again.

PR #496, opened 2015-07-27
This should be a min rather than max when compute the resize ratio

YongzhiWang:
This should be a min rather than max. For example, a 2000 * 1000 image with resize option 200*200, using max, the result will be 400*200. But when the user gives a 200*200 contraint, the expected result should be 200*100.


IanChilds:
It's not at all clear that this should be a min rather than a max. If the aspect ratio of your image is very high, and you use min, then you will get a much lower quality image that you might expect. 

Currently, we care more about image quality than memory use, so we'll be leaving it as it is for the moment. The best solution would be to make this configurable.

YongzhiWang:
I think it is the about the clearance of the interface design. When I call the ResizeOption with a 200 * 200 constraint, I expect it to be a image which is totally within this range without surprise. But now, the behavior is strange and I do not think other image library keep this behavior! In terms of the image quality, I believe you should leave the decision to the user of this function because they know what size of the image they want.

PR #563, opened 2015-08-20
added a New Application Name using Fresco

vipulasri:
I have added a new name to applications using Fresco Library


vipulasri:
I have replaced by app using Universal Image Loader with Fresco and its pretty awesome. Thank you Facebook for a such a great Open Source.

PR #586, opened 2015-08-31
Set default placeholder scaleType if none is provided

tbruceyu:
This is a bug fix if we want to use the setPlaceHolderImage in GenericDraweeHierarchy to set a place holder image.
Sometimes, we need to set the placeHolder in DraweeHierarchy. But if
we never set the placeHolder before, the setPlaceholderImage would not
set the scaleType. This fix added a default scale type when we set the
place holder image in code.

### DIFF
--- a/drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
+++ b/drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
@@ -151,6 +151,7 @@ public class GenericDraweeHierarchy implements SettableDraweeHierarchy {
   private final int mFailureImageIndex;
   private final int mControllerOverlayIndex;
 
+  private ScaleType mPlaceHolderScaleType;
   private RoundingParams mRoundingParams;
 
   GenericDraweeHierarchy(GenericDraweeHierarchyBuilder builder) {
@@ -173,9 +174,13 @@ public class GenericDraweeHierarchy implements SettableDraweeHierarchy {
         mRoundingParams,
         mResources,
         placeholderImageBranch);
+    mPlaceHolderScaleType = builder.getPlaceholderImageScaleType();
+    if (mPlaceHolderScaleType == null) {
+        mPlaceHolderScaleType = GenericDraweeHierarchyBuilder.DEFAULT_SCALE_TYPE;
+    }
     placeholderImageBranch = maybeWrapWithScaleType(
-        placeholderImageBranch,
-        builder.getPlaceholderImageScaleType());
+            placeholderImageBranch,
+            mPlaceHolderScaleType);
     mPlaceholderImageIndex = numLayers++;
 
     // actual image branch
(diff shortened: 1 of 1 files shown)


tbruceyu:
OK. Thanks for your reminding. I had signed up the facebook code. It's great to join your community. If other things I should do, please let me know. Thanks.

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
```diff
@@ -640,6 +645,7 @@ public void setPlaceholderImage(Drawable drawable) {
       drawable = getEmptyPlaceholderDrawable();
     }
     drawable = maybeApplyRoundingBitmapOnly(mRoundingParams, mResources, drawable);
+    maybeWrapWithScaleType(drawable, mPlaceHolderScaleType);
```
plamenko: This line should not be necessary.

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
```diff
@@ -151,6 +151,7 @@ public void draw(Canvas canvas) {
   private final int mFailureImageIndex;
   private final int mControllerOverlayIndex;
 
+  private ScaleType mPlaceHolderScaleType;
```
plamenko: No need for this to be a member variable.

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
```diff
@@ -173,9 +174,13 @@ public void draw(Canvas canvas) {
         mRoundingParams,
         mResources,
         placeholderImageBranch);
+    mPlaceHolderScaleType = builder.getPlaceholderImageScaleType();
+    if (mPlaceHolderScaleType == null) {
+        mPlaceHolderScaleType = GenericDraweeHierarchyBuilder.DEFAULT_SCALE_TYPE;
+    }
```
plamenko: If we decide to do this, it would be better to change [GenericDraweeHierarchyBuilder] (https://github.com/facebook/fresco/blob/920539d9443bc3979fbd838ecd64d3356b735ee0/drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchyBuilder.java#L83) than to do it here.

However, I don't think we should do this at all. This change will cause a ScaleTypeDrawable to be always created for each placeholder even though in some cases this is not necessary. E.g. color drawables (including a default placeholder which is just a transparent drawable) do not need any scaling because scaling does not affect them. This will impose an unnecessary memory overhead.

If you are specifying the placeholder programmatically after the hierarchy is being built, and you don't have the placeholder available at the time the hierarchy is being built, then you can achieve the same behavior by doing `GenericDraweeHierarchyBuilder.setPlaceholderImage(null, someScaleType)`. This will ensure that a ScaleTypeDrawable for the placeholder image gets created.

plamenko:
@tbruceyu thank you for this pull request!
I labeled it as `please-change` as I'd like to discuss whether this is a necessary or even a desirable change. If you strongly feel that `GenericDraweeHierarchyBuilder.setPlaceholderImage(null, someScaleType)` is not a convenient solution, then we can consider doing this, but it would be better to do it in `GenericDraweeHierarchyBuilder`.
Also, please keep in mind that this is a breaking change as we are now changing the default behavior someone might have relied on.

tbruceyu:
@plamenko Very thanks for your careful review. I think this issue could be closed now. GenericDraweeHierarchyBuilder.setPlaceholderImage(null, someScaleType) is OK for my issue.

plamenko:
Awesome! Btw, please do not hesitate to send more pull-requests!

Issue #736, opened 2015-11-06
OkHttp 404 responses written to cache anyway


NextApp:
![47 pic](https://cloud.githubusercontent.com/assets/4578812/10999128/450832da-84d5-11e5-9337-ced3cd2b281a.jpg)

LouisCAD:
You should check `response.isSuccess()` before trying to use your response

erikandre:
Closing as the issue has already been fixed

PR #745, opened 2015-11-10
Fix for OkHttp backend caching 404 responses as images

andreidiaconu:
Issue link: https://github.com/facebook/fresco/issues/736
Unless the response is successful, I am treating it as a normal exception.

According to OkHttp documentation, onResponse is called if comunication with the server resulted in a response (meaning no IO problems): http://square.github.io/okhttp/2.x/okhttp/com/squareup/okhttp/Callback.html#onResponse-com.squareup.okhttp.Response-
This means that we need to handle non-200 responses in onResponse, since a 404 from the server is still a response. The issue describes fresco caching 404 responses, for instance.


IanChilds:
This is really great, thanks @andreidiaconu!!!

shercoder:
What about apps that fetch images from Cloudinary where if the image does not exist so 404 is thrown but the response comes with a default image (404 image) to display. However @andreidiaconu solution prevents from getting these default 404 images.

I came across this issue using React Native's `Image.getSize` method which uses `DataSource` to get the image but end up getting `Failure` because of the solution above.

Should this be something configurable, where you can allow non-200 responses to be cached based on the configuration?

lambdapioneer:
@shercoder: You can create you own `NetworkFetcher` implementation and provide that to the `ImagePipeline.Config`. With that you'd have all freedom to handle these special cases.

I feel that most applications want to do proper error handling of any non-200 response that comes from the network stack on the product level. For instance, often one does not want such an image staying in the cache for a transient issue.

Issue #511, opened 2015-08-04
Can't use drawable with colorFilter as placeholder when roundAsCircle is true


radon:
I'm trying to use a drawable with a color filter. If i set the rounding params i get the placeholder without color filter, if i remove them it works correctly.
This is the code i'm using:

```
       SimpleDraweeView simpleDraweeView = (SimpleDraweeView) findViewById(R.id.draweeview);

        Drawable drawable = getResources().getDrawable(R.drawable.placeholder);
        drawable.setColorFilter(getResources().getColor(R.color.green), PorterDuff.Mode.SRC_IN);

        GenericDraweeHierarchyBuilder builder =
                new GenericDraweeHierarchyBuilder(getResources());
        GenericDraweeHierarchy hierarchy = builder
                .setPlaceholderImage(drawable)
                .setRoundingParams(
                        new RoundingParams()
                                .setBorder(getResources().getColor(R.color.gray), 1)

                ).build();


        simpleDraweeView.setHierarchy(hierarchy);
        simpleDraweeView.setImageURI(Uri.parse("http://png-1.findicons.com/files/icons/1072/face_avatars/300/i04.png"));
```

Thanks :)

PR #752, opened 2015-11-11
When using RoundingParams, placeholder image no longer loses Paint properties

andreidiaconu:
Fix for this issue: https://github.com/facebook/fresco/issues/511

When constructing a new RoundedBitmapDrawable from a source BitmapDrawable, any modifications done on the Paint of the original BitmapDrawable get lost because only the Bitmap is transmitted along.
I have created another constructor for RoundedBitmapDrawable that accepts a Paint object, so that mPaint can be created as a copy of it.
I also wrote a new test that reproduces issue #511 - that is now passing and was not before.

### DIFF
--- a/drawee/src/main/java/com/facebook/drawee/drawable/RoundedBitmapDrawable.java
+++ b/drawee/src/main/java/com/facebook/drawee/drawable/RoundedBitmapDrawable.java
@@ -49,7 +49,7 @@ public class RoundedBitmapDrawable extends BitmapDrawable
   private final Path mPath = new Path();
   private boolean mIsPathDirty = true;
   /** True if this rounded bitmap drawable will actually do anything. */
-  private final Paint mPaint = new Paint(Paint.ANTI_ALIAS_FLAG);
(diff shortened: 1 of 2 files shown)


drawee/src/main/java/com/facebook/drawee/drawable/RoundedBitmapDrawable.java
```diff
@@ -57,7 +57,17 @@
   private @Nullable TransformCallback mTransformCallback;
 
   public RoundedBitmapDrawable(Resources res, Bitmap bitmap) {
+    this(res, bitmap, null);
+  }
+
+  public RoundedBitmapDrawable(Resources res, Bitmap bitmap, Paint paint){
     super(res, bitmap);
+    if (paint == null){
```
massimocarli: Please fix code conventions. 
```
if (paint == null) {
  // code
} else {
  // code
}
```

IanChilds:
This is already imported - don't worry about updating the commit. Will be in the next version.

PR #775, opened 2015-11-17
Add CircleProgressBar

cazador4:
Example:

```java
CircleProgressBarDrawable progressBar = new CircleProgressBarDrawable();
progressBar.setBackgroundColor(context.getResources().getColor(R.color.grey));
progressBar.setColor(context.getResources().getColor(R.color.red));
GenericDraweeHierarchyBuilder builder = new GenericDraweeHierarchyBuilder(context.getResources());
builder.setFadeDuration(30).setProgressBarImage(progressBar).build();
```

### DIFF
--- a/samples/contrib/com/facebook/drawee/drawable/CircleProgressBarDrawable.java
+++ b/samples/contrib/com/facebook/drawee/drawable/CircleProgressBarDrawable.java
@@ -1,0 +1,42 @@
+package com.facebook.drawee.drawable;
+
+import android.graphics.Canvas;
+import android.graphics.Paint;
+import android.graphics.Rect;
+import android.graphics.RectF;
+
+import com.facebook.drawee.drawable.ProgressBarDrawable;
+
+public class CircleProgressBarDrawable extends ProgressBarDrawable {
+    private final Paint mPaint = new Paint(Paint.ANTI_ALIAS_FLAG);
+    private int mLevel = 0;
+    private int maxLevel = 10000;
+
+
+    @Override
+    protected boolean onLevelChange(int level) {
+        mLevel = level;
+        invalidateSelf();
+        return true;
+    }
+
+    @Override
+    public void draw(Canvas canvas) {
+        if (getHideWhenZero() && mLevel == 0) {
+            return;
+        }
+        drawBar(canvas, maxLevel, getBackgroundColor());
+        drawBar(canvas, mLevel, getColor());
+    }
+
+    private void drawBar(Canvas canvas, int level, int color) {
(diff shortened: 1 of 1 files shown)


hanhmh1203:
hi, progressbar change size follow image
ex: progressbar show eclipse on image. how to show progressbar always circle ???

plzen:
@hanhmh1203 I rewrote this method:

```
private void drawBar(Canvas canvas, int level, int color) {
        Rect bounds = getBounds();
        int minLength = Math.min(bounds.width(), bounds.height());
        float progressSize = minLength * 0.2f;
        RectF rectF = new RectF(
                bounds.centerX() - progressSize / 2,
                bounds.centerY() - progressSize / 2,
                bounds.centerX() + progressSize / 2,
                bounds.centerY() + progressSize / 2);
        mPaint.setColor(color);
        mPaint.setStyle(Paint.Style.STROKE);
        mPaint.setStrokeWidth(6);
        if (level != 0)
            canvas.drawArc(rectF, 0, (float) (level * 360 / maxLevel), false, mPaint);
    }
```

balazsbalazs:
Hey @cazador4,

Thx for the pull request, it's awesome.
However, we discussed it internally and we'd prefer not to include it in the core library, rather include it in a separate samples/contrib directory.

Our reasons are the following:
1. We want to keep the core library as small and as simple as possible.
2. You can do a similar thing using AutoRotateDrawable as the progress bar image
3. The class is very simple and there's a chance users will need a custom implementation of  (eg stroke width, circle style, etc).

That said I think the code is **very** useful so we'd love to have it because I'm pretty sure people do need it. We also have a few other Pull Requests which are similar.
So what I suggest: let's put this under /samples/contrib/ directory. (/samples/contrib/com/facebook/drawee/drawable/CircleProgressBarDrawable.java)

What do you think?

cazador4:
Hey Balazs, It's ok for me.
I've changed that.

Thanks!
Daniel

balazsbalazs:
Ah, cool, thx for this!

rafaelekol:
@cazador4  I tried to implement your [CircleProgressBarDrawable](https://github.com/facebook/fresco/blob/2ed3a9dd56e14d7211f328c9952d114118b6598f/samples/contrib/com/facebook/drawee/drawable/CircleProgressBarDrawable.java), but instead of circle shape I get Oval. What could be the problem? Maybe because my DraweeView is not square?

cazador4:
Hi @rafaelekol, maybe that will be the problem.
You could try the @plzen answer, changing the rectF.

PR #812, opened 2015-12-04
Add description about the executor

Kroderia:
Add description about the executor in dataSource subscribe

### DIFF
--- a/docs/02-datasources-datasubscribers.md
+++ b/docs/02-datasources-datasubscribers.md
@@ -115,5 +115,16 @@ DataSubscriber dataSubscriber =
 dataSource.subscribe(dataSubscriber, executor);
 ```
 
+### For executor
+
+```java
+    Executor executor = new Executor() {
+        @Override
+        public void execute(Runnable command) {
+            command.run();
+        }
+    }
+```
+
 If you want to deviate from the example above and assign the `CloseableReference` to another variable somewhere else, you can. Just be sure to [follow the rules](closeable-references.html).
 


plamenko:
Hi, I think you are missing the point of executors. This is not how executors are supposed to be implemented/used. The purpose of executors is to execute runnables on a specific thread and with specific policy.
We already provide several [executors] (https://github.com/facebook/fresco/tree/0f3d52318631f2125e080d2a19f6fa13a31efb31/fbcore/src/main/java/com/facebook/common/executors) and one should carefully chose which one to be used. The one that you implemented is actually `CallerThreadExecutor` and should be used with great caution as it executes runnables on the caller's thread.
If you need to do any UI stuff from your callback, you must use `UiThreadExecutorService` as Android view system is not thread safe and is only to be accessed from the main thread (the UI thread).
If you need to do some db access, file writes or anything else that is a slow operation, this should not be done either with CallerThread nor with the UiThread executors, but with one of the background thread executors. See [DefaultExecutorSupplier.forBackgroundTasks] (https://github.com/facebook/fresco/blob/0f3d52318631f2125e080d2a19f6fa13a31efb31/imagepipeline/src/main/java/com/facebook/imagepipeline/core/DefaultExecutorSupplier.java) for example implementation.

Kroderia:
Thanks, I just found nothing about the executor in the doc and try to make it. Fresco's document just not clear enough.

plamenko:
No problem, thank you for pointing this out. I just updated the docs explaining executors. Let me know if that helps.

plamenko:
Btw, Executors are not really Fresco specific, this is a well known Java concept. But I do agree that it is worth mentioning them in the docs.
Also, please see:
https://docs.oracle.com/javase/tutorial/essential/concurrency/executors.html
http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executor.html

Kroderia:
Oh my..I am really newbie for this. Thanks for your patience.

PR #829, opened 2015-12-10
Update current_version in _config.yml

kunny:
Updated current_version from 0.7.0 to 0.8.1.


tyronen:
We left it at 0.7.0 on purpose because of issues syncing to jCenter.

PR #875, opened 2016-01-02
Upgrade to OkHttp 3.

swankjesse:


tyronen:
Hi Jesse. Thanks for the request! Two things--

1. We do need you to sign the CLA, and
2. This release is numbered as RC1. If it's only a release candidate, I'd prefer waiting until it has graduated from it.

I didn't see any release notes for 3.0 - is there a description anywhere of what's changed?

swankjesse:
CLA is in-process. My company’s already signed it & I’m in the process of getting my username on the list.

Changelog is here:
https://github.com/square/okhttp/blob/master/CHANGELOG.md

None of the APIs that impact Fresco will be impacted by the differences between RC1 and 3.0.0 final. (I’m the project’s maintainer.)

tyronen:
We've talked about it a bit here at Facebook. We totally understand that, with a new API, you wouldn't want libraries like ours blocking its adoption. And we'd like to see apps using our libraries to get the benefits of your upgrade sooner rather than later.

ReactNative uses both Fresco and OkHttp directly, and of course we have internal code using all three. A change of this scale takes a certain amount of effort, mostly in testing. We'd prefer to go through this exercise only once rather than twice to make absolutely sure we get it right. For that reason I'd still like to wait on this until it reaches the 3.0 milestone.

That doesn't mean we're halting on this - we've started to test 3.0.0-RC1 internally and will let you know if there are problems with it. But this particular PR will have to wait a bit, sorry.

balazsbalazs:
As @tyronen  mentioned the biggest work here to migrate all of our internal dependencies to OKHttp3.

The quickest and safest way to get this in would be to create a separate okhttp3 imagepipeline-backend.
So add a new directory fresco/imagepipeline-backends/imagepipeline-okhttp3 which will produce a new artifact: imagepipeline-okhttp3.aar

This has an added benefit that we're not introducing a breaking change, people can still use the new version of Fresco with the old OKHttp. (Although I guess for OKHttp's adoption point of view this is not a bonus :) )

The downside is that we're duplicating some code but we're talking about two small classes here.

What do you think?

swankjesse:
@balazsbalazs a separate imagepipeline-backend is a good idea. Can I also deprecate the old one? That makes the duplication less offensive.

balazsbalazs:
Yeah, that makes a lot of sense and it also makes people (and us ;) ) update faster.

lzy26fate:
I reference old version,write a new version,you can look at https://github.com/lzy26fate/imagepipeline-okhttp3

balazsbalazs:
@swankjesse - any update on this?
Or alternatively @lzy26fate would you like to create the pull request?

swankjesse:
I’ll take a look at this tonight.

swankjesse:
(thanks for your patience!)

swankjesse:
Ready to go.

swankjesse:
Ping!

balazsbalazs:
Hi Jessee,

We already imported your PR but until we do a new release it's not going to show up. We're currently working on better github integration - that's one of the main reason why there was no new release for a while.

As of today:
The source of truth is our internal repo and we do a manual sync from internal to github at every release (similar to AOSP). So your pull request is not going to be visible until we do a new release.

What we're working on:
We want the internal and github repos to be completely in sync. This means that once we accept your pull request it will automatically show up in github. There are a number of issues we need to solve to achieve this (external CI, sync script, warnings for leaking internal info, etc) but we're very close.

Hopefully we'll finish this shortly and also make a new version with OkHTTP3.

swankjesse:
Sounds good.

PR #909, opened 2016-01-13
add fitx and fity types in scale type

yangkai3:
image y-dimension will be show entirely

### DIFF
--- a/drawee/src/main/java/com/facebook/drawee/drawable/ScalingUtils.java
+++ b/drawee/src/main/java/com/facebook/drawee/drawable/ScalingUtils.java
@@ -85,8 +85,14 @@ public class ScalingUtils {
     FOCUS_CROP,
 
     /**
-     * keep height dimension fix exactly and aspect ratio is preserved.
-     * It maybe only show center part of width.
+     * Scales the image so that the x dimension fits exactly and aspect ratio is preserved.
+     * Image is vertically centered in the parent.
+     */
+    FIT_X,
+
+    /**
+     * Scales the image so that the y dimension fits exactly and aspect ratio is preserved.
+     * Image is horizontally centered in the parent.
      */
     FIT_Y
   }
@@ -199,17 +205,20 @@ public class ScalingUtils {
         transform.postTranslate((int) (dx + 0.5f), (int) (dy + 0.5f));
         break;
 
+      case FIT_X:
+        scale = scaleX;
+        dx = parentBounds.left;
+        dy = parentBounds.top + (parentHeight - childHeight * scale) * 0.5f;
+        transform.setScale(scale, scale);
+        transform.postTranslate((int) (dx + 0.5f), (int) (dy + 0.5f));
+        break;
+
       case FIT_Y:
         scale = scaleY;
-
-        float localScaleX = scaleY;
-        if (childWidth * scale < parentWidth) {
-          localScaleX = scaleX;
-        }
-
-        dx = parentBounds.left + (parentWidth - childWidth * localScaleX) * 0.5f;
-        transform.setScale(localScaleX, scale);
-        transform.postTranslate((int) (dx + 0.5f), 0);
+        dx = parentBounds.left + (parentWidth - childWidth * scale) * 0.5f;
+        dy = parentBounds.top;
+        transform.setScale(scale, scale);
+        transform.postTranslate((int) (dx + 0.5f), (int) (dy + 0.5f));
         break;
 
       default:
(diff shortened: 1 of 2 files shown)


yangkai3:
Contributor License Agreement signed

drawee/src/main/java/com/facebook/drawee/drawable/ScalingUtils.java
```diff
@@ -83,11 +82,18 @@
      * It is guaranteed that the focus point will be visible and centered as much as possible.
      * If the focus point is set to (0.5f, 0.5f), result will be equivalent to CENTER_CROP.
      */
-    FOCUS_CROP
+    FOCUS_CROP,
+
+    /**
+     * keep height dimension fix exactly and aspect ratio is preserved.
```
plamenko:
```
/**
 * Scales the image so that the y dimension fits exactly and aspect ratio is preserved.
 * Image is horizontally centered in the parent.
 */
```

drawee/src/main/res/values/attrs.xml
```diff
@@ -26,6 +26,7 @@
       <enum name="centerInside" value="5" />
       <enum name="centerCrop" value="6" />
       <enum name="focusCrop" value="7" />
+      <enum name="fixY" value="8" />
```
plamenko: `fitY`, not `fixY`. On other places too.

plamenko:
Thanks for sending a pull request. Please see my comments. Also, it would be nice to have the equivalent logic for `FIT_X`.

drawee/src/main/java/com/facebook/drawee/drawable/ScalingUtils.java
```diff
@@ -193,6 +199,19 @@ public static Matrix getTransform(
...
+      case FIT_Y:
+        scale = scaleY;
+
+        float localScaleX = scaleY;
+        if (childWidth * scale < parentWidth) {
+          localScaleX = scaleX;
+        }
+
+        dx = parentBounds.left + (parentWidth - childWidth * localScaleX) * 0.5f;
+        transform.setScale(localScaleX, scale);
```
plamenko: This is incorrect. In order to preserve the aspect ratio, image must be scaled uniformly by x and y.

```
transform.setScale(scale, scale);
```

drawee/src/main/java/com/facebook/drawee/drawable/ScalingUtils.java
```diff
@@ -193,6 +199,19 @@ public static Matrix getTransform(
         transform.postTranslate((int) (dx + 0.5f), (int) (dy + 0.5f));
         break;
 
+      case FIT_Y:
+        scale = scaleY;
+
+        float localScaleX = scaleY;
```
plamenko: Remove those 4 lines. We need to preserve the aspect ratio which means we should not stretch the image, even if it doesn't fill the parent.

yangkai3:
hi plamenko ：
 i modify my code and add fitX type.

plamenko:
Looks good thanks!

Btw, just a notice that I am doing some refactoring in order to allow providing of a custom scale type without having to modify the Drawee code. We will import your pull request, but the code will likely be refactored once my change gets in. My refactor shouldn't affect the existing clients so no need to worry about that. I just wanted to let you know.

yangkai3:
It will be a useful function to let user provide custom scale type!

oprisnik:
Sorry for the delay! 
The way scale types work has changed and you can now specify custom scale types.
Unfortunately, your PR has merge conflicts because of this.

The scale types in ScalingUtils have to be updated to the new format where each scale type is a separate class.
For example, something like that should work:

```java
private static class ScaleTypeFitX extends AbstractScaleType {

  public static final ScaleType INSTANCE = new ScaleTypeFitX();

  @Override
  public void getTransformImpl(
      Matrix outTransform,
      Rect parentRect,
      int childWidth,
      int childHeight,
      float focusX,
      float focusY,
      float scaleX,
      float scaleY) {
    float scale, dx, dy;
    scale = scaleX;
    dx = parentRect.left;
    dy = parentRect.top + (parentRect.height() - childHeight * scale) * 0.5f;
    outTransform.setScale(scale, scale);
    outTransform.postTranslate((int) (dx + 0.5f), (int) (dy + 0.5f));
  }
}
```

In addition, you need to also update `GenericDraweeHierarchyInflater` to support the new scale types from XML. Finally, it would be great to have tests for the new scale types similar to the ones we have for existing types.

It would be great if you could update your PR to the new format. If you'd like we could also implement a new version ourselves when we have some time.

kirwan:
I'm going to close this PR as it's not been updated for a while and we now offer custom scale types. If you still want to update it, please do.

oprisnik:
I have added an example for custom scale types to our showcase sample app that adds fitX and fitY, see https://github.com/facebook/fresco/commit/92da47e95664414f0d442dbd2fbe88daaa563384

Issue #912, opened 2016-01-14
Not able to set hotspot on "overlayImage" drawable set in xml 


sagartrehan:
I have set "app:overlayImage" drawable on SimpleDraweeView in xml layout for adding selector ripple effect on top of SimpleDraweeView. Setting "overlayImage" attribute correctly showing ripple effect on image but hot spot is always showing from center. To solve this I need the overlayImage drawable that I had set in xml resource but I didn't find any method to get overlayImage drawable programmatically.

If I prepare GenericDraweeHierarchy programmatically then I can prepare overlayImage drawable and can set touch listener hot spot to it. Code snippet:

SimpleDraweeView image = (SimpleDraweeView) findViewById(R.id.image);
GenericDraweeHierarchyBuilder builder = new GenericDraweeHierarchyBuilder(mContext.getResources());
Drawable overlayImage = getDrawable(R.drawable.image_selector);
builder.setOverlay(overlayImage);

image.setHierarchy(builder.build());

if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
view.setOnTouchListener(new View.OnTouchListener() {
@TargetApi(Build.VERSION_CODES.LOLLIPOP)
@Override
public boolean onTouch(View v, MotionEvent event) {
if (overlayImage != null) {
overlayImage.setHotspot(event.getX(), event.getY());
}
return false;
}
});
}

But I do not want to create GenericDraweeHierarchy programmatically as I had already set different attributes in xml and do not want to change it and set programmatically. Please suggest any method to get "overlayImage" drawable that I had set in xml layout resource.

Please let me know if I can add hot spot in any other way

sagartrehan:
Let me know if more details needed

plamenko:
I think there is an easy fix to that. As of API 21 Drawable got a new method: ` void setHotspot (float x, float y)`. We just need to override this method in our `ForwardingDrawable` and `ArrayDrawable` and forward the call to the children.
We might want to make this feature optional though, because I am concerned by performance implications of passing each touchevent to the whole hierarchy of drawables. Anyhow, this should be easy to fix on Fresco side.

plamenko:
[...]

PR #967, opened 2016-01-31
Fix for #912 Not able to set hotspot on "overlayImage" drawable set i…

sagartrehan:
…n xml

- Overridden setHotspot(float x, float y)  method in ForwardingDrawablea and ArrayDrawable for passing touch event to drawee layers to make ripple hot spot work with SimpleDraweeView "overlayImage" attribute.

### DIFF
--- a/drawee/src/main/java/com/facebook/drawee/drawable/ArrayDrawable.java
+++ b/drawee/src/main/java/com/facebook/drawee/drawable/ArrayDrawable.java
@@ -376,7 +376,8 @@ public class ArrayDrawable extends Drawable
   @Override
   @TargetApi(Build.VERSION_CODES.LOLLIPOP)
   public void setHotspot(float x, float y) {
-    for (Drawable drawable : mLayers) {
+    for (int i = 0; i < mLayers.length; i++) {
+      Drawable drawable = mLayers[i];
       if (drawable != null) {
         drawable.setHotspot(x, y);
       }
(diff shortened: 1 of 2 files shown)


drawee/src/main/java/com/facebook/drawee/drawable/ArrayDrawable.java
```diff
@@ -370,4 +372,14 @@ public void getRootBounds(RectF bounds) {
       bounds.set(getBounds());
     }
   }
+
+  @Override
+  @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+  public void setHotspot(float x, float y) {
+    for (Drawable drawable : mLayers) {
```
kirwan: Please change this foreach to a for loop over the array as with every other method here which loops through the drawables.

sagartrehan: OK I will update code with for loop instead of foreach and update pull request

sagartrehan:
I have updated the Pull-Request and used for loop instead of foreach for iterating our layer of drawables.

plamenko:
Thank you!

plamenko:
Btw, the reason we use for loop instead of for-each loop is because for-each automatically creates a new iterator and this allocation may cause GC to happen which is something we don't want.

sagartrehan:
Thanks I was not aware of this difference.

oprisnik:
@facebook-github-bot import

facebook-github-bot-0:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1500865480221734/int_phab) to review.

PR #971, opened 2016-02-02
Add a link to app that uses Fresco

eldk:
Hello,

Should I Add a link out of Playstore ?

The main page of the app is there : http://www.opalesurfcasting.net/legislation/lamaille_-_une_application_android_pour_ne_pas_oublier_la_maille_article3004.html

Greatings,

Eric


tyronen:
So far we have used Play Store links for apps and site links for libraries. Unless there's a compelling reason not to, we'd rather keep it this way.

tyronen:
To save time we added it directly.

eldk:
Thank you.

PR #999, opened 2016-02-17
Update MultiPointerGestureDetector.java

levianye:
fix double tap zoom image bug


levianye:
it seems double tap to zoom image will shake so I fix it and it works well in my project

plamenko:
Hi! Thank you for your pull request. However, I do have a working double-tap to zoom in/out code and no changes are necessary to the gesture detectors. I will push the code to the github soon.

PR #1028, opened 2016-03-01
Create WrapContentDraweeView

Ilya-Gazman:
Added support for WRAP_CONTENT option. Tested in my app and it works pretty well, allows me to use both adjustViewBounds and WRAP_CONTENT.

The only problem is that it don't show a proper preview. Perhaps someone will be able to assist here.

### DIFF
--- a/drawee/src/main/java/com/facebook/drawee/view/WrapContentDraweeView
+++ b/drawee/src/main/java/com/facebook/drawee/view/WrapContentDraweeView
@@ -1,0 +1,125 @@
+package com.facebook.drawee.view;
+
+import android.content.Context;
+import android.content.res.Resources;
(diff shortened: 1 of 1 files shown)


Ilya-Gazman:
Oh, nice. 

Ok I signed up.

plamenko:
Thank you for the pull-request. However, this is something we deliberately decided not to do. Read [here](why) http://frescolib.org/docs/wrap-content.html#_. We will reconsider this decision, and if we decide to support `wrap_content`, this would have to be done on the `DraweeView` level for flexibility reasons. There are many custom views that extend `DraweeView` directly (instead of `SimpleDraweeView`) and we would need to support this for these other views too.

PR #1056, opened 2016-03-14
Add Vimeo to the list of orgs using Fresco

kevinzetterstrom:
Adding Vimeo to the list of organizations/apps that use Fresco


oprisnik:
Great, thank you very much!

Issue #1011, opened 2016-02-23
Allow to disable crossfade


AlexBlokh:
When crossfading `placeholder` and `downloaded image` - both images changes alpha channel from `0..1` and from `1..0` 
TransitionDrawable has an option `.setCrossFadeEnabled(boolean enabled);`
This option will lead to a smoother transition between `placeholder drawable (colour drawable)` and `downloaded drawable`
it should be done in `FadeDrawable.class`

aagnes:
I don't understand what you mean? Is that a question or a feature you want fresco library to have?

AlexBlokh:
![zqfl8](https://cloud.githubusercontent.com/assets/4045375/13280267/348ad4cc-dadd-11e5-9f66-a3e3d761138c.gif)
on the left - transition with crossfade, on the right - without

aagnes:
It would be great if you want to contribute and have a pull request after.

AlexBlokh:
got it

AlexBlokh:
spent half a day and still can't build the project :O

plamenko:
Instead of `placeholderImage`, sat the `backgroundImage` to black.

PR #1070, opened 2016-03-19
CrossFade disabling feature

AlexBlokh:
#1011 issue


drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchyBuilder.java
```diff
@@ -373,7 +375,7 @@ public GenericDraweeHierarchyBuilder setOverlays(List<Drawable> overlays) {
    * @return modified instance of this builder
    */
   public GenericDraweeHierarchyBuilder setOverlay(Drawable overlay) {
-    mOverlays = Arrays.asList(overlay);
+    mOverlays = Collections.singletonList(overlay);
```
AlexBlokh: Suggested by the IDE

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchyBuilder.java
```diff
@@ -347,7 +349,7 @@ public GenericDraweeHierarchyBuilder setBackgrounds(List<Drawable> backgrounds)
    * @return modified instance of this builder
    */
   public GenericDraweeHierarchyBuilder setBackground(Drawable background) {
-    mBackgrounds = Arrays.asList(background);
+    mBackgrounds = Collections.singletonList(background);
```
AlexBlokh: Suggested by the IDE

plamenko: Unrelated change. Please make a separate pull request.

plamenko:
There are a couple of things going on here.
* FadeDrawable already supports independently fading each layer. Therefore, no changes are necessary to it.
* As for the GenericDraweeHierarchy changes, it is not clear what is the fading semantics with crossfade disabled.

Can you give us more context on what are you trying to achieve actually? If you do not wish placeholder to be faded out when fading to the actual image, you can set it as a background instead. That way it will stay faded in. We will be able to help once you provide more context on what behavior would you liek to achieve.

AlexBlokh:
As I mentioned in #1011 I want to achieve the same behaviour as TransitionDrawalbe.crossfadeEnabled(false).

When new image will be crossfaded in, the foreground one would become invisible, reducing overdrawing

I'll review the feedback and update the pull request

plamenko:
Thanks! I believe you can actually achieve what you want without any changes to Fresco. See my comment on the issue.

PR #1098, opened 2016-03-31
fix DecodeProducer & LocalFetchProducer

huangjunkun:
1. fix DecodeProducer encodeImage has no metedata；   
   leading to decode source(big) Bitmap Object .
2. improve LocalFetchProducer's logic.

### DIFF
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java
@@ -122,9 +122,6 @@ public class DecodeProducer implements Producer<CloseableReference<CloseableImag
               ImageRequest request = producerContext.getImageRequest();
               if (mDownsampleEnabledForNetwork ||
                   !UriUtil.isNetworkUri(request.getSourceUri())) {
-                if (!EncodedImage.isMetaDataAvailable(encodedImage)) {
-                  encodedImage.parseMetaData();
-                }
                 encodedImage.setSampleSize(DownsampleUtil.determineSampleSize(
                     request, encodedImage));
               }
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java
@@ -122,6 +122,9 @@ public class DecodeProducer implements Producer<CloseableReference<CloseableImag
               ImageRequest request = producerContext.getImageRequest();
               if (mDownsampleEnabledForNetwork ||
                   !UriUtil.isNetworkUri(request.getSourceUri())) {
+                if (!EncodedImage.isMetaDataAvailable(encodedImage)) {
+                  encodedImage.parseMetaData();
+                }
                 encodedImage.setSampleSize(DownsampleUtil.determineSampleSize(
                     request, encodedImage));
               }
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/HttpUrlConnectionNetworkFetcher.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/HttpUrlConnectionNetworkFetcher.java
@@ -51,57 +51,57 @@ public class HttpUrlConnectionNetworkFetcher extends BaseNetworkFetcher<FetchSta
   @Override
   public void fetch(final FetchState fetchState, final Callback callback) {
     final Future<?> future = mExecutorService.submit(
-        new Runnable() {
-          @Override
-          public void run() {
-              HttpURLConnection connection = null;
-              Uri uri = fetchState.getUri();
-              String uriString = uri.toString();
-              InputStream is;
-              try {
-                  connection = createGetConnection(uriString);
-                  int redirectCount = 0;
-                  while (connection.getResponseCode() / 100 == 3 && redirectCount < MAX_REDIRECT_COUNT) {
-                      uriString = connection.getHeaderField("Location");
-                      connection.disconnect();
-                      connection = createGetConnection(uriString);
-                      ++redirectCount;
-                  }
-                  if (redirectCount > MAX_REDIRECT_COUNT) {
-                      callback.onFailure(new Exception("RedirectCount more than " + MAX_REDIRECT_COUNT));
-                      return;
-                  }
-                  is = connection.getInputStream();
-                  callback.onResponse(is, -1);//connection.getContentLength()
-              } catch (Exception e) {
-                  callback.onFailure(e);
-              } finally {
-                  if (connection != null) {
-                      connection.disconnect();
-                  }
-              }
-          }
-        });
-    fetchState.getContext().addCallbacks(
-        new BaseProducerContextCallbacks() {
-          @Override
-          public void onCancellationRequested() {
-            if (future.cancel(false)) {
-              callback.onCancellation();
+      new Runnable() {
+        @Override
+        public void run() {
+          HttpURLConnection connection = null;
+          Uri uri = fetchState.getUri();
+          String uriString = uri.toString();
+          InputStream is;
+          try {
+            connection = createGetConnection(uriString);
+            int redirectCount = 0;
+            while (connection.getResponseCode() / 100 == 3 && redirectCount < MAX_REDIRECT_COUNT) {
+              uriString = connection.getHeaderField("Location");
+              connection.disconnect();
+              connection = createGetConnection(uriString);
+              ++redirectCount;
+            }
+            if (redirectCount >= MAX_REDIRECT_COUNT) {
+              callback.onFailure(new Exception("RedirectCount more than " + MAX_REDIRECT_COUNT));
+              return;
+            }
+            is = connection.getInputStream();
+            callback.onResponse(is, -1);//connection.getContentLength()
+          } catch (Exception e) {
+            callback.onFailure(e);
+          } finally {
+            if (connection != null) {
+                connection.disconnect();
             }
           }
-        });
+        }
+      });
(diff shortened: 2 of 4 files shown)


imagepipeline/src/main/java/com/facebook/imagepipeline/producers/LocalFileFetchProducer.java
```diff
@@ -33,9 +34,9 @@ public LocalFileFetchProducer(
 
   @Override
   protected EncodedImage getEncodedImage(final ImageRequest imageRequest) throws IOException {
-    return getEncodedImage(
-        new FileInputStream(imageRequest.getSourceFile().toString()),
-        (int) imageRequest.getSourceFile().length());
+      File sourceFile = imageRequest.getSourceFile();
+      return getEncodedImage(sourceFile,
+              new FileInputStream(imageRequest.getSourceFile().getPath()), (int) sourceFile.length());
```
kirwan: You've still left a reference to `imageRequest.getSourceFile()`.

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java
```diff
@@ -122,6 +122,9 @@ public void run(EncodedImage encodedImage, boolean isLast) {
               ImageRequest request = producerContext.getImageRequest();
               if (mDownsampleEnabledForNetwork ||
                   !UriUtil.isNetworkUri(request.getSourceUri())) {
+                if (!EncodedImage.isMetaDataAvailable(encodedImage)) {
```
kirwan: This is currently done in `AddImageTransformMetaDataProducer` which is included in the producer sequence for local images so shouldn't be needed here.

huangjunkun: @kirwan 
AddImageTransformMetaDataProducer is too late possibly. 
DecodeProducer's encodedImage.setSampleSize may be invalid.
And this problem  is not always occur.

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/LocalFileFetchProducer.java
huangjunkun: That has some problem？
I think it is OK.

huangjunkun:
Anybody go on doing code review  ? @facebook-github-bot

kirwan:
First, thanks for your PR. We really appreciate contributions from others.

It would help if you could explain what problem you're trying to fix though. I don't see a link to an issue this should resolve and you haven't explained the scenario where there is a problem without these changes.

Also, you've added some HTTP changes which don't relate to the title or description (and should probably be in a separate PR) and some of your changes are incorrectly formatted (in particular, we use two-space indents).

If you can explain the purpose of the changes we can better review them. It would particularly help if you could include an example `ImageRequest` and `ImagePipelineConfig` which benefit from this.

kirwan:
Also, have you signed our CLA? It looks like the Github Bot has not updated the PR to say you have.

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/HttpUrlConnectionNetworkFetcher.java
```diff
@@ -47,37 +54,33 @@ public void fetch(final FetchState fetchState, final Callback callback) {
...
-                if (nextUriString == null || nextScheme.equals(scheme)) {
+                  connection = createGetConnection(uriString);
+                  int redirectCount = 0;
+                  while (connection.getResponseCode() / 100 == 3 && redirectCount < MAX_REDIRECT_COUNT) {
+                      uriString = connection.getHeaderField("Location");
+                      connection.disconnect();
+                      connection = createGetConnection(uriString);
+                      ++redirectCount;
+                  }
+                  if (redirectCount > MAX_REDIRECT_COUNT) {
```
duchuyctlk: How can `redirectCount` be larger than `MAX_REDIRECT_COUNT`? When `redirectCount` is equal to `MAX_REDIRECT_COUNT`, the statements inside `while` block will not be executed so its value will not increase.

facebook-github-bot-5:

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/LocalFetchProducer.java
```diff
@@ -105,20 +105,35 @@ protected EncodedImage getByteBufferBackedEncodedImage(
...
-    long javaFree = Math.min(javaMax - javaUsed, 8 * ByteConstants.MB);
-    if (mDecodeFileDescriptorEnabledForKitKat && inputStream instanceof FileInputStream &&
-        javaMax >= 64 * javaFree) {
-      return getInputStreamBackedEncodedImage(new File(inputStream.toString()), length);
-    } else {
-      return getByteBufferBackedEncodedImage(inputStream, length);
+    private boolean isMemoryEnough() {
+        Runtime runTime = Runtime.getRuntime();
+        long javaMax = runTime.maxMemory();
+        long javaUsed = runTime.totalMemory() - runTime.freeMemory();
```
plamenko: These calls might be expensive. I would be very cautious with that.

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/LocalFileFetchProducer.java
plamenko: It's just a coding style, since you already have `sourceFile = imageRequest.getSourceFile()`, you can just do `sourceFile.getPath()` instead of `imageRequest.getSourceFile().getPath()`.

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/HttpUrlConnectionNetworkFetcher.java
plamenko: The problem is that the code above doesn't distinguish whether we exited the loop because we reached limit, or because we got a valid code.

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/HttpUrlConnectionNetworkFetcher.java
```diff
@@ -47,37 +54,33 @@ public void fetch(final FetchState fetchState, final Callback callback) {
...
               InputStream is;
               try {
-                URL url = new URL(uriString);
-                connection = (HttpURLConnection) url.openConnection();
-                nextUriString = connection.getHeaderField("Location");
-                nextScheme = (nextUriString == null) ? null : Uri.parse(nextUriString).getScheme();
-                if (nextUriString == null || nextScheme.equals(scheme)) {
+                  connection = createGetConnection(uriString);
+                  int redirectCount = 0;
+                  while (connection.getResponseCode() / 100 == 3 && redirectCount < MAX_REDIRECT_COUNT) {
```
plamenko:
```
while (connection.getResponseCode() / 100 == 3 &&
    ++redirectCount <= MAX_REDIRECT_COUNT) {
  ...
}
```

Notice the preincrement in the check **after**  the status code check. Now after the loop one of those two things is true:
* redirectCount <= MAX_REDIRECT_COUNT  && code != 3xx
* redirectCount > MAX_REDIRECT_COUNT && code == 3xx

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/HttpUrlConnectionNetworkFetcher.java
plamenko: And obviously, this needs a unit test

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/HttpUrlConnectionNetworkFetcher.java
```diff
@@ -47,37 +54,33 @@ public void fetch(final FetchState fetchState, final Callback callback) {
...
               } catch (Exception e) {
-                callback.onFailure(e);
-                break;
+                  callback.onFailure(e);
               } finally {
-                if (connection != null) {
-                  connection.disconnect();
-                }
+                  if (connection != null) {
+                      connection.disconnect();
```
plamenko: Style: use 2 spaces and not tabs

plamenko:
It would be good to make two separate pull requests, one for LocalFetcher, the other for NetworkFetcher, as these two seems to be unrelated.

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/HttpUrlConnectionNetworkFetcher.java
huangjunkun: OK 
Fix bug：if (redirectCount >= MAX_REDIRECT_COUNT)  { ... }

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/LocalFetchProducer.java
huangjunkun: I think that is little optimizing.

facebook-github-bot-1:
@huangjunkun updated the pull request.

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java
huangjunkun: @kirwan  OK, you're right.I got it wrong.

facebook-github-bot-7:
@huangjunkun updated the pull request.

facebook-github-bot-4:
@huangjunkun updated the pull request.

PR #1124, opened 2016-04-17
Added support for synchronous disk cache check

chethann:
Added a support to check in disk cache synchronously. The call will block the caller thread. 
This is useful when you want to manually check if an image is present in disk cache inside a async callback, this prevents nesting of callbacks.

### DIFF
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/cache/BufferedDiskCache.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/cache/BufferedDiskCache.java
@@ -125,6 +125,32 @@ public class BufferedDiskCache {
   }
 
   /**
+   * Performs disk cache check synchronously.
+   * @param key
+   * @return true if the key is found in disk cache else false
+   */
+  public boolean syncDiskCheck(final CacheKey key){
+    if(containsSync(key)){
+      return true;
+    }
+    EncodedImage result = mStagingArea.get(key);
+    if (result != null) {
+      result.close();
+      FLog.v(TAG, "Found image for %s in staging area", key.toString());
+      mImageCacheStatsTracker.onStagingAreaHit();
+      return true;
+    } else {
+      FLog.v(TAG, "Did not find image for %s in staging area", key.toString());
+      mImageCacheStatsTracker.onStagingAreaMiss();
+      try {
+        return mFileCache.hasKey(key);
+      } catch (Exception exception) {
+        return false;
+      }
+    }
+  }
+
+  /**
    * Performs key-value look up in disk cache. If value is not found in disk cache staging area
    * then disk cache read is scheduled on background thread. Any error manifests itself as
    * cache miss, i.e. the returned task resolves to null.
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/cache/BufferedDiskCache.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/cache/BufferedDiskCache.java
@@ -116,7 +116,7 @@ public class BufferedDiskCache {
    * @return true if the key is found in disk cache else false
    */
   public boolean syncDiskCheck(final CacheKey key) {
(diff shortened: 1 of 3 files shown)


kirwan:
The change looks good but I have some small comments to add. Also, it would be great if you could add a unit test or two for the new methods.

imagepipeline/src/main/java/com/facebook/imagepipeline/cache/BufferedDiskCache.java
```diff
@@ -125,6 +125,32 @@ public Boolean call() throws Exception {
   }
 
   /**
+   * Performs disk cache check synchronously.
+   * @param key
+   * @return true if the key is found in disk cache else false
+   */
+  public boolean syncDiskCheck(final CacheKey key){
+    if(containsSync(key)){
```
kirwan: Space missing before opening bracket.

imagepipeline/src/main/java/com/facebook/imagepipeline/cache/BufferedDiskCache.java
```diff
@@ -125,6 +125,32 @@ public Boolean call() throws Exception {
...
+      return true;
+    }
+    EncodedImage result = mStagingArea.get(key);
```
kirwan: This looks exactly like the task content inside `containsAsync(CacheKey)` so it be great if you could change that to call this method rather than having the duplication.

imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipeline.java
```diff
@@ -391,6 +391,28 @@ public boolean isInBitmapMemoryCache(final ImageRequest imageRequest) {
 
   /**
    * Returns whether the image is stored in the disk cache.
+   * The check is made synchronously
```
kirwan: This doesn't look complete. The Javadoc of `isInDiskCacheSync(ImageRequest)` is more explicit about blocking the thread so please add the same or a similar clarification here.

imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipeline.java
```diff
@@ -391,6 +391,28 @@ public boolean isInBitmapMemoryCache(final ImageRequest imageRequest) {
...
+   * @param uri the uri for the image to be looked up.
+   * @return true if the image was found in the disk cache, false otherwise.
+   */
+  public boolean isInDiskCacheSync(final Uri uri) {
+    return isInDiskCacheSync(ImageRequest.fromUri(uri));
+  }
+
+  /**
+   * Performs disk cache check synchronously. It is not recommended to use this
+   * unless u know what exactly you are doing. Disk cache check is a costly operation,
```
kirwan: "u" -> "you"

chethann:
@kirwan I have made changes based on your comments. Please have a look!

imagepipeline/src/main/java/com/facebook/imagepipeline/cache/BufferedDiskCache.java
```diff
@@ -125,6 +111,18 @@ public Boolean call() throws Exception {
   }
 
   /**
+   * Performs disk cache check synchronously.
+   * @param key
+   * @return true if the key is found in disk cache else false
+   */
+  public boolean syncDiskCheck(final CacheKey key) {
+    if(containsSync(key)){
```
kirwan: You're still missing a space before the opening bracket. Otherwise this looks good to me.

chethann: My bad. Fixed it now

kirwan:
Thanks for this. We currently have an issue with out import script but this should be imported in the next few days.

kirwan:
@facebook-github-bot import

PR #1139, opened 2016-04-21
Add launch post as first post in blog

kirwan:
This gives the new Fresco blog its first post, the original lauch from code.facebook.com.

It looks a little something like this:

screen shot 2016-04-21 at 5 27 46 pm


oprisnik:
Awesome :)

PR #1141, opened 2016-04-22
Update repo link in footer

andrewimm:
"redex" => "fresco" :)

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -17,7 +17,7 @@
         <a class="footerLink" href="https://twitter.com/fbOpenSource" target="_blank">Twitter</a>
       </div>
       <div class="footerSection rightAlign">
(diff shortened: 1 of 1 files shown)


andrewimm:
Also, I love that I was able to do this without leaving github.com

kirwan:
Thanks very much. The new website just went live yesterday and we hadn't noticed that.

matthewjohnston4:
Oh, actually, I'll suggest an improvement here. Because `ghrepo` is defined in [_config.yml](https://github.com/facebook/fresco/blob/gh-pages/_config.yml#L13) it becomes available to use in templates like this footer as {{ site.ghrepo }}, so you should be able to change that to:

```
 <a class="footerLink" href="https://github.com/{{ site.ghrepo }}" target="_blank">Contribute to this project on GitHub</a>
```

kirwan:
OK, I've just merged the PR so I'll change that separately.

PR #1155, opened 2016-04-25
Fix navigation links

kirwan:
This fixes the navigation to both the external links and the internal links when the site is viewed from http://facebook.github.io/fresco/


oprisnik:
This also fixes the broken GitHub link :)

PR #1165, opened 2016-04-27
fix logo url in getting started.

ndurell:
The current url in the docs does not resolve to anything. Maybe it changed?


oprisnik:
Thanks!

PR #1168, opened 2016-04-28
Updated docs to refer to OkHttp 3 instead of 2

oprisnik:
This is similar to #1078. Since OkHttp 3 is now available, we should prefer it over v2.


facebook-github-bot-3:
@oprisnik updated the pull request.

Issue #22, opened 2015-03-28
Shared element transition


jorgemf:
Shared element transition doesn't seem to work.

Tested in a motorola device with android 5.0

tyronen:
Can you give us more details? What exactly did you try to do, and what happened instead?

jorgemf:
In the new android version, lollipop, we can use an ImageView as a shared element transition between activities. In that case the image make a smoth transition from one activity to the next one. When using the SimpleDraweeView it just disappear.

The code to test it should be something like this:

theme of the app:

```xml
	<style name="AppTheme" parent="Theme.AppCompat.Light.NoActionBar">
		<item name="android:windowContentTransitions">true</item>
		<item name="android:windowAllowEnterTransitionOverlap">true</item>
		<item name="android:windowAllowReturnTransitionOverlap">true</item>
	</style>
```

Activity1 Layout.xml:

```xml
<RelativeLayout
	xmlns:android="http://schemas.android.com/apk/res/android"
	android:layout_width="match_parent"
	android:layout_height="match_parent">

	<com.facebook.drawee.view.SimpleDraweeView
		android:id="@+id/image"
		android:layout_width="@dimen/size_1"
		android:layout_height="@dimen/size_1"
		android:layout_alignParentLeft="true"
		android:layout_alignParentTop="true"/>

</RelativeLayout>
```

Activity2 Layout.xml:

```xml
<RelativeLayout
	xmlns:android="http://schemas.android.com/apk/res/android"
	android:layout_width="match_parent"
	android:layout_height="match_parent">

	<com.facebook.drawee.view.SimpleDraweeView
		android:id="@+id/image"
		android:layout_width="@dimen/size_2"
		android:layout_height="@dimen/size_2"
		android:layout_alignParentRight="true"
		android:layout_alignParentBottom="true"
		android:transitionName="image_transition"/>

</RelativeLayout>
```

The code to start the second activity:

```java
	Intent intent = new Intent(activity1, Activity2.class);
	ActivityOptions options = ActivityOptions.makeSceneTransitionAnimation(activity1, simpleDraweeView1, "image_transition");
	activity1.startActivity(intent, options.toBundle());
```

This code should make a transition between the activities where the image moves and resizes from the top left corner of the activity 1 to the bottom right corner of the activity 2. I have tested a similar code loading images from the network and it does not display the animation.

jorgemf:
This is the same issue: https://github.com/facebook/fresco/issues/99

[...]

PR #1217, opened 2016-05-13
Fix shared elements transition on Lollipop

Aohayou:
Hello,
As referenced in https://github.com/facebook/fresco/issues/22, there is currently a problem for shared elements transitions that occurs on Lollipop devices but not on Marshmallow (As far as I know).

Here is the problem, illustrated by your sample app "transition":
before
I'm performing a click on the start transition button, then going back. The transition from the first to the second activity shows resize issues, and the transition back to the first activity doesn't display anything.

after
With the same actions, we have the same visual result as with a device running Marshmallow.

Since ChangeBounds is animating the bounds of the the view, we need to react to the onSizeChanged callback, which is the fix I've done.

### DIFF
--- a/drawee/src/main/java/com/facebook/drawee/view/DraweeView.java
+++ b/drawee/src/main/java/com/facebook/drawee/view/DraweeView.java
@@ -238,7 +238,8 @@ public class DraweeView<DH extends DraweeHierarchy> extends ImageView {
     super.onSizeChanged(w, h, oldw, oldh);
     Drawable drawable = getTopLevelDrawable();
     if (drawable != null) {
-      drawable.setBounds(0, 0, w, h);
+      drawable.setBounds(getPaddingLeft(), getPaddingTop(),
+              w - getPaddingLeft() - getPaddingRight(), h - getPaddingTop() - getPaddingBottom());
     }
   }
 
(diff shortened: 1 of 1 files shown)


drawee/src/main/java/com/facebook/drawee/view/DraweeView.java
```diff
@@ -234,6 +234,15 @@ protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
   }
 
   @Override
+  protected void onSizeChanged(int w, int h, int oldw, int oldh) {
+    super.onSizeChanged(w, h, oldw, oldh);
+    Drawable drawable = getTopLevelDrawable();
+    if (drawable != null) {
+      drawable.setBounds(0, 0, w, h);
```
plamenko: Unfortunately this doesn't account for potential padding set on the view. We need to account for that too.

plamenko:
Thanks for making this fix! Please make sure that it works if padding is set on the view and then we can get this merged.

facebook-github-bot-9:
@Aohayou updated the pull request.

Aohayou:
Thank you for your feedback. Here is a version that accounts for potential padding set on the view.

kirwan:
@facebook-github-bot import

facebook-github-bot-6:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/567549926756788/int_phab) to review.

kirwan:
@Aohayou Thanks so much for this PR but another came very soon after which I means I don't think is necessary anymore. See #1225.

PR #1225, opened 2016-05-17
Drawee scaletype transition

dmitrychistyakov:
As mentioned in #22 ChangeImageTransform for DraweeViews is not working correctly, when ScaleType is changing during transition. This pull request resolve this issue by added custom DraweeTransition, which animates ScaleType transition correctly

InterpolatingScaleType is changed by adding start and end values of parentBounds. In pervious version it used parameter parentBounds, which changing every animation frame and it caused wrong animation behaviour.

Transition sample is changed due to use DraweeTransition. Also background was added in sample image, to show how animation works


drawee/src/main/java/com/facebook/drawee/drawable/ScalingUtils.java
```diff
@@ -365,15 +365,19 @@ public void getTransformImpl(
 
     private final ScaleType mScaleTypeFrom;
     private final ScaleType mScaleTypeTo;
-    private final float[] mMatrixValuesFrom = new float[9];
-    private final float[] mMatrixValuesTo = new float[9];
+    private final Rect mBoundsFrom;
```
plamenko: Let's make this two bounds fields @Nullable

drawee/src/main/java/com/facebook/drawee/drawable/ScalingUtils.java
```diff
@@ -415,10 +427,16 @@ public Matrix getTransform(
         int childHeight,
         float focusX,
         float focusY) {
-      mScaleTypeFrom.getTransform(transform, parentBounds, childWidth, childHeight, focusX, focusY);
-      transform.getValues(mMatrixValuesFrom);
-      mScaleTypeTo.getTransform(transform, parentBounds, childWidth, childHeight, focusX, focusY);
-      transform.getValues(mMatrixValuesTo);
+      if (mMatrixValuesFrom == null) {
+        mMatrixValuesFrom = new float[9];
+        mScaleTypeFrom.getTransform(transform, mBoundsFrom, childWidth, childHeight, focusX, focusY);
```
plamenko: Let's use `parentBounds` in case `mBoundsFrom` is null. Same for `mBoundsTo` below.

`Rect boundsFrom = (mBoundsFrom != null) ? mBoundsFrom : parentBounds;`
`Rect boundsTo = (mBoundsTo != null) ? mBoundsTo : parentBounds;`

drawee/src/main/java/com/facebook/drawee/drawable/ScalingUtils.java
```diff
@@ -365,15 +365,19 @@ public void getTransformImpl(
...
+    private final Rect mBoundsFrom;
+    private final Rect mBoundsTo;
+    private float[] mMatrixValuesFrom;
+    private float[] mMatrixValuesTo;
     private final float[] mMatrixValuesInterpolated = new float[9];
 
     private float mInterpolatingValue;
 
-    public InterpolatingScaleType(ScaleType scaleTypeFrom, ScaleType scaleTypeTo) {
+    public InterpolatingScaleType(ScaleType scaleTypeFrom, ScaleType scaleTypeTo, Rect boundsFrom, Rect boundsTo) {
```
plamenko: I'd have both constructors:

```
    public InterpolatingScaleType(ScaleType scaleTypeFrom, ScaleType scaleTypeTo) {
      this(scaleTypeFrom, scaleTypeTo, null, null);
    }
```

drawee/src/main/java/com/facebook/drawee/drawable/ScalingUtils.java
```diff
@@ -415,10 +427,16 @@ public Matrix getTransform(
         int childHeight,
         float focusX,
         float focusY) {
-      mScaleTypeFrom.getTransform(transform, parentBounds, childWidth, childHeight, focusX, focusY);
-      transform.getValues(mMatrixValuesFrom);
-      mScaleTypeTo.getTransform(transform, parentBounds, childWidth, childHeight, focusX, focusY);
-      transform.getValues(mMatrixValuesTo);
+      if (mMatrixValuesFrom == null) {
```
plamenko: I am a bit concerned about this as there may be some correctness issues. For example, if the child dimensions change in the middle of the animation this would cause a display issue. Let's just leave this computation unconditionally as it was before.

Context:
Child bounds indeed may change in the middle of the animation. Fresco allows specifying low-res and high-res image when building a Drawee controller, and once the high-res image arrives it will automatically replace the previously displayed low-res image. Low-res and high-res images are exactly the same except that the low-res image is a downscaled version of the high-res one. The content and the aspect ratio are exactly the same, but width and height are not. With a proper scale type (e.g. center_crop or fit_center) this swap will go completely smooth, but for that to happen scale type always needs to recompute the transform.

drawee/src/main/java/com/facebook/drawee/view/DraweeTransform.java
```diff
@@ -0,0 +1,83 @@
...
+ * smoothly.
+ */
+public class DraweeTransform extends Transition {
+
+  private static final String PROPNAME_BOUNDS = "draweeTransform:bounds";
+
+  private final ScalingUtils.ScaleType mFromScale;
+  private final ScalingUtils.ScaleType mToScale;
+
+  public DraweeTransform(ScalingUtils.ScaleType fromScale, ScalingUtils.ScaleType toScale) {
```
plamenko: Let's keep `fromScale` and `toScale` in `transitionValues`.

drawee/src/main/java/com/facebook/drawee/view/DraweeTransform.java
```diff
@@ -0,0 +1,83 @@
...
+        draweeView.getHierarchy().setActualImageScaleType(mToScale);
+      }
+    });
+
+    return animator;
+  }
+
+  private void captureValues(TransitionValues transitionValues) {
+    if (transitionValues.view instanceof GenericDraweeView) {
+      transitionValues.values.put(PROPNAME_BOUNDS, new Rect(0, 0, transitionValues.view.getWidth(), transitionValues.view.getHeight()));
```
plamenko:
```
GenericDraweeView draweeView = (GenericDraweeView) transitionValues.view;
transitionValues.values.put(SCALE_TYPE, draweeView.gethierarchy().getActualImageScaleType());
...
```

drawee/src/main/java/com/facebook/drawee/view/DraweeTransform.java
```diff
@@ -0,0 +1,83 @@
...
+
+@TargetApi(Build.VERSION_CODES.KITKAT)
+/**
+ * This Transition animates changes of {@link GenericDraweeView} between two ScaleTypes
+ *
+ * In combination with ChangeBounds, DraweeTransform allows GenericDraweeViews
+ * that change size, shape, or {@link ScalingUtils.ScaleType} to animate contents
+ * smoothly.
+ */
+public class DraweeTransform extends Transition {
```
plamenko: For ease of usage, let's make a static factory method here that will create the necessary TransitionSet:

```
public static TransitionSet createTransitionSet() {
  TransitionSet transitionSet = new TransitionSet();
  transitionSet.addTransition(new ChangeBounds());
  transitionSet.addTransition(new DraweeTransform());
  return transitionSet;
}
```

samples/transition/src/main/java/com/facebook/samples/transitions/DestinationActivity.java
```diff
@@ -14,12 +14,38 @@
...
+
+    setupEnterTransition();
+    setupReturnTransition();
+  }
+
+  private void setupEnterTransition() {
+    TransitionSet transitionSet = new TransitionSet();
+    transitionSet.addTransition(new ChangeBounds());
+    transitionSet.addTransition(new DraweeTransform(ScalingUtils.ScaleType.CENTER_CROP, ScalingUtils.ScaleType.CENTER_INSIDE));
+    getWindow().setSharedElementEnterTransition(transitionSet);
```
plamenko: Just:

```
getWindow().setSharedElementEnterTransition(DraweeTransform.createTransitionSet());
```

samples/transition/src/main/java/com/facebook/samples/transitions/DestinationActivity.java
```diff
@@ -14,12 +14,38 @@
...
+  }
+
+  private void setupReturnTransition() {
+    TransitionSet transitionSet = new TransitionSet();
+    transitionSet.addTransition(new ChangeBounds());
+    transitionSet.addTransition(new DraweeTransform(ScalingUtils.ScaleType.CENTER_INSIDE, ScalingUtils.ScaleType.CENTER_CROP));
```
plamenko: As above, just:

getWindow().setSharedElementEnterTransition(DraweeTransform.createTransitionSet());

plamenko:
Looks amazing, just a couple of comments to address.

plamenko:
I am accepting this in advance as there are only a couple small comments to be applied.

drawee/src/main/java/com/facebook/drawee/view/DraweeTransform.java
dmitrychistyakov: This is how I try to do it first. But it doesn't work unfortunately.

The reason is - when transition from Source Activity to Destination occurs, it doesn't use Source view itself, it uses Source view parameters such as bounds and apply it to view in Destination Activity. Then, it smoothly change these parameters for view.
Thats why, in captureStartValues and in captureEndValues we'll got same ScaleType (as in Destination Activity) and animation will be broken.

For ImageView, which works without additional parameters, Android passes right values in captureStartValues, because it saves ImageView matrix at the captureSharedElementState step along with other parameters and then restore this matrix and apply it to ImageView before captureStartValues. See ActivityTransitionCoordinator KEY_SCALE_TYPE usage.

As I understand, we don't have such possibility, thats why constructor parameters used.Maybe do you have other suggestions to do it?

facebook-github-bot-1:
@dmitrychistyakov updated the pull request.

drawee/src/main/java/com/facebook/drawee/view/DraweeTransform.java
plamenko: Hmm, that's really strange. If that's the case, how does `transitionValues.view.getHeight()` work? Let me double check what is really happening here because I don't see the point of `captureStartValues` if the view passed in is not the source view.

Anyhow, if we don't find a workaround for this I am okay with just leaving it as is. But let's double check first.

plamenko: Okay I was looking into Android framework code and it makes me sad. I found this [patch](https://android.googlesource.com/platform/frameworks/base/+/080443b%5E!/) where they added transitions support for ImageView. It basically boils down to hardcoding bunch of Imageview specific stuff in the core framework classes. There appears to be no extensible way of doing this for an arbitrary custom view.

Let's then keep it like this until we find a better solution.

drawee/src/main/java/com/facebook/drawee/view/DraweeTransform.java
plamenko: Btw, let's call this class `DraweeTransition` instead of `DraweeTransform`

facebook-github-bot-0:
@dmitrychistyakov updated the pull request.

kirwan:
This looks fantastic. Thanks for the PR. Let me know if you have any more amendments, otherwise I'll import it so we can merge to master.

dmitrychistyakov:
@kirwan Thank you, I think, I've done with this PR (if you don't have comments to be applied of course).

I'm researching possibility declare DraweeTransition in xml like this
```
<?xml version="1.0" encoding="utf-8"?>
<transitionSet xmlns:android="http://schemas.android.com/apk/res/android">
 <changeBounds/>
 <transition class="com.facebook.drawee.view.DraweeTransition" fromScaleType="centerCrop" toScaleType="centerInside"/>
</transitionSet>
```

which will make usage of DraweeTransition simpler, but I think it's would be better to make separate PR for this, if it's possible

kirwan:
@facebook-github-bot import

kirwan:
@dmitrychistyakov I'm curious what the reason was for removing the transparency from `test_image.png`. I presume that's not necessary.

dmitrychistyakov:
@kirwan sample image with transparent bg looks similar with ScaleType.Crop and ScaleType.CenterInside in example (just little difference in size). Thats why you can't see how matrix is changing. I added bg to show how actually CenterCrop animates to CenterInside. But it's not necessary of cource, i can revert it.

plamenko:
Thanks @dmitrychistyakov! I stepped through the framework code and I can confirm that what you said is exactly right. The framework copies just some of the properties from the source view to the destination view and then passes the destination view instance to `Transition.captureStartValues`. It then restores back the original properties of the destination view and calls `Transition.captureEndValues`. This is a design flaw of the Android framework as it prevents any custom properties to be properly captured. We will have to go with the current solution until this gets improved on Android side.

As for adding XML support for `DraweeTransition`, that would be amazing too! As you said, a separate PR would be better than doing it here. Let's ship this!

plamenko:
Btw, about making transition more obvious, I changed the destination scale type to FIT_CENTER and added a backgroundImage which is just a light green color. The transition is very visible.
screen shot 2016-05-19 at 15 57 40
screen shot 2016-05-19 at 15 57 49

kirwan:
@dmitrychistyakov I've run this on a device now and it looks brilliant. Thanks so much.

There's no need to make any more changes, as I had to run some formatting anyway so I've added the background as @plamenko suggested.

PR #1242, opened 2016-05-26
[Fresco] Fix long press in ZoomableDraweeView

plamenko:
Tested that before this diff a false long-press would be triggered if user taps while the image is not loaded yet. Tested that this does not happen with this fix, and that long-press properly gets detected once the image gets loaded.

### DIFF
--- a/samples/zoomable/src/main/java/com/facebook/samples/zoomable/ZoomableDraweeView.java
+++ b/samples/zoomable/src/main/java/com/facebook/samples/zoomable/ZoomableDraweeView.java
@@ -274,6 +274,16 @@ public class ZoomableDraweeView extends DraweeView<GenericDraweeHierarchy> {
       FLog.v(getLogTag(), "onTouchEvent: %d, view %x, handled by the super", a, this.hashCode());
       return true;
     }
+    // None of our components reported that they handled the touch event. Upon returning false
+    // from this method, our parent won't send us any more events for this gesture. Unfortunately,
+    // some componentes may have started a delayed action, such as a long-press timer, and since we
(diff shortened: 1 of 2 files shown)


samples/zoomableapp/src/main/java/com/facebook/samples/zoomableapp/MyPagerAdapter.java
```diff
@@ -49,7 +53,13 @@ public int getCount() {
...
 	}
-}
-
-
 
+	private GestureDetector.SimpleOnGestureListener createTapListener(final int position) {
+		return new GestureDetector.SimpleOnGestureListener() {
+			@Override
+			public void onLongPress(MotionEvent e) {
+				Log.d("plamenko", "onLongPress: " + position);
```
balazsbalazs: haha, replace the tag :)

samples/zoomable/src/main/java/com/facebook/samples/zoomable/ZoomableDraweeView.java
```diff
@@ -274,6 +274,16 @@ public boolean onTouchEvent(MotionEvent event) {
       FLog.v(getLogTag(), "onTouchEvent: %d, view %x, handled by the super", a, this.hashCode());
       return true;
     }
+    // None of the our components reported that they handled the touch event. Upon returning false
```
balazsbalazs: Type: remove the "the" (None of **the** our components...)

balazsbalazs:
That's a very nice catch! Mega Like!
(Please fix the nits, otherwise it's good to go)

facebook-github-bot-9:

PR #1243, opened 2016-05-26
[Fresco] Properly handle requestDisallowInterceptTouchEvent

plamenko:
If the current gesture performed on our ZoomableDraweeView haven't caused the transform to be corrected, we request our parent to not intercept events as we are going to hndle them. If on the other hand the transform got corrected, it means we tried to scroll past the bounds and we will let our parent intercept the events in case it wants to.

As a consequence we can zoom-in, pan the image around, and once we reach the edge, the parent will take over and perform paging, as the user would expect. With this fix I hope we solved issues with ZoomableDraweeView within another scrollable containers such as ViewPager, etc.

### DIFF
--- a/samples/zoomable/src/main/java/com/facebook/samples/zoomable/DefaultZoomableController.java
+++ b/samples/zoomable/src/main/java/com/facebook/samples/zoomable/DefaultZoomableController.java
@@ -74,6 +74,7 @@ public class DefaultZoomableController
   private final Matrix mActiveTransformInverse = new Matrix();
   private final float[] mTempValues = new float[9];
   private final RectF mTempRect = new RectF();
+  private boolean mWasTransformCorrected;
 
   public static DefaultZoomableController newInstance() {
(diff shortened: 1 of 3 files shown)


balazsbalazs:
Ah, I get it now why we need the ViewPager in the sample app :)

balazsbalazs:
This is very cool! I've checked the implementation in the Android source for requestDisallowInterceptTouchEvent for ViewGroup, ScrollView and the AbsListView - this should work.
I'd live to test this with the Facebook app - but it looks good.

AWESOME!

balazsbalazs:
@facebook-github-bot import

facebook-github-bot-1:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1733629793588247/int_phab) to review.

facebook-github-bot-9:

PR #1246, opened 2016-05-26
[Fresco] Always reserve space for a background and an overlay

plamenko:
This achieves two things:
* It allows us to dynamically change backgrounds and overlays
* It saves memory as we removed 5 instance integers and added only 2 instance null-references (two null elements in the array of the fade drawable)

An additional change is that multiple backgrounds are from now on deprecated. Context:
This improvement required a small interface change between GenericDraweeHierarchy and its builder. In particular, GenericDraweeHierarchy now has only one layer dedicated for a background. Multiple backgrounds are still supported via builder, but this gets emulated by wrapping them with an ArrayDrawable. In theory this should work fine, with one caveat. Rounding no longer works for multiple backgrounds. This would be relatively easy to support, but given that multiple backgrounds are rarely used feature (if ever), instead of introducing additional complexity we decided to deprecate it instead.
In case where there is only one or no background, everything should work as it was before.

Test Plan: unit tests, demo sample app

### DIFF
--- a/drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
+++ b/drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
@@ -72,7 +72,7 @@ import static com.facebook.drawee.drawable.ScalingUtils.ScaleType;
  * <li> All branches except the actual image branch are optional (placeholder, failure, retry,
  * progress bar). If some branch is not specified it won't be created. Index in FadeDrawable will
  * still be reserved though.
- * <li> If overlays and/or backgrounds are specified, they are added to the same fade drawable, and
+ * <li> If overlays and/or background are specified, they are added to the same fade drawable, and
  * are always being displayed.
  * <li> ScaleType and Matrix transformations will be added only if specified. If both are
  * unspecified, then the branch for that image is attached to FadeDrawable directly. Matrix
@@ -85,6 +85,14 @@ import static com.facebook.drawee.drawable.ScalingUtils.ScaleType;
  */
 public class GenericDraweeHierarchy implements SettableDraweeHierarchy {
 
+  private static final int BACKGROUND_IMAGE_INDEX = 0;
+  private static final int PLACEHOLDER_IMAGE_INDEX = 1;
+  private static final int ACTUAL_IMAGE_INDEX = 2;
+  private static final int PROGRESS_BAR_IMAGE_INDEX = 3;
+  private static final int RETRY_IMAGE_INDEX = 4;
+  private static final int FAILURE_IMAGE_INDEX = 5;
+  private static final int OVERLAY_IMAGES_INDEX = 6;
+
   private final Drawable mEmptyActualImageDrawable = new ColorDrawable(Color.TRANSPARENT);
 
   private final Resources mResources;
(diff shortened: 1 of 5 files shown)


drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
```diff
@@ -85,6 +85,14 @@
...
 public class GenericDraweeHierarchy implements SettableDraweeHierarchy {
 
+  private static final int BACKGROUND_IMAGE_INDEX = 0;
+  private static final int PLACEHOLDER_IMAGE_INDEX = 1;
+  private static final int ACTUAL_IMAGE_INDEX = 2;
+  private static final int PROGRESS_BAR_IMAGE_INDEX = 3;
+  private static final int RETRY_IMAGE_INDEX = 4;
+  private static final int FAILURE_IMAGE_INDEX = 5;
+  private static final int OVERLAY_IMAGES_INDEX = 6;
+
```
massimocarli: What about using @IntDef annotation here?

plamenko: What's the benefit?

massimocarli: Oh, I didn't realized they where private. Anyway it would add type safety-ish.

balazsbalazs: I like adding @IntDef - it's a nice touch.

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
```diff
@@ -555,26 +544,9 @@ public void setProgressBarImage(int resourceId, ScaleType scaleType) {
...
   }
 
-  /**
-   * Sets a new background image at the specified index.
-   *
-   * This method will throw if the given index is out of bounds.
-   *
-   * @param drawable background image
-   */
-  public void setBackgroundImage(int index, @Nullable Drawable drawable) {
```
balazsbalazs: as we discussed IRL: you just introduced this method a few days ago, right?

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
```diff
@@ -585,14 +557,10 @@ public void setBackgroundImage(@Nullable Drawable drawable) {
    * @param drawable background image
    */
   public void setOverlayImage(int index, @Nullable Drawable drawable) {
-    // Note: we do not have mOverlaysIndex to reduce the memory footprint of this class.
```
balazsbalazs: This two lines of comments are still relevant: we do rely on overlays being the last ones.

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchyBuilder.java
```diff
@@ -581,34 +582,36 @@ public GenericDraweeHierarchyBuilder setActualImageColorFilter(@Nullable ColorFi
...
     return this;
   }
 
   /**
-   * Gets the backgrounds.
+   * Gets the background.
    */
-  public @Nullable List<Drawable> getBackgrounds() {
-    return mBackgrounds;
+  public @Nullable Drawable getBackground() {
```
balazsbalazs: Ok, this is definitely a breaking change. Which is fine but let's put this in the description of the PR.
"From now on Fresco only supports one background instead of multiple ones. We've seen very little usage of multiple backgrounds and we can have better performance if we disallow multiple backgrounds"

samples/demo/src/main/res/layout/activity_main.xml
```diff
@@ -110,8 +110,6 @@
       <com.facebook.drawee.view.SimpleDraweeView
         android:id="@+id/alpha_webp"
         style="@style/drawee"
-        fresco:backgroundImage="@color/transparent"
```
balazsbalazs: ? Why did you remove these?

balazsbalazs:
All in all a great change but let's do a more in-depth description.
We're basically removing support for multiple backgrounds with this commit - let's emphasize that. I would change the commit name to "Remove multiple background support" and then explain that we're doing this for perf reasons.

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
plamenko: Pushing back again :) It duplicates the definitions without a good reason. Those are index constants, not enum values.

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
plamenko: Correct. It should be safe to remove it.

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchyBuilder.java
plamenko: This getter is in `GenericDraweeHierarchyBuilder`, and is most likely used exclusively by `GenericDraweeHierarchy` constructor. No one should really call this directly.

samples/demo/src/main/res/layout/activity_main.xml
plamenko: Because we no longer need to preallocate space for background/overlay in order to change it dynamically later on, and more importanitly we no longer need to allocate this useless transparent drawable.
Pure win :)

plamenko:
We are not really dropping support for multiple backgrounds. There has been some internal interface changes between drawee hierarchy and its builder, but the builder setters are still there.

There is a breaking part though, which is that rounding won't work for backgrounds if there is more than one. This is easy to fix, but I didn't bother because I am reasonably confident that multiple backgrounds is rarely used if ever. I admit that was a bit ov overengineering on my side. Time to fix it :)

facebook-github-bot-6:

PR #1287, opened 2016-06-09
[Fresco][RFC] Add TEMP_BITMAP rounding mode

plamenko:
Summary:
This mode doesn't suffer any limitations as other modes, but it requires extra memory.

Test Plan:
Rounding sample app


plamenko:
This is how it looks with this patch:

screen shot 2016-06-09 at 19 03 22

plamenko:
I still need to figure out how to reuse the bitmap efficiently instead of allocating in draw.

kirwan:
I'm closing our very old PRs from internal people. We may still pursue this but right now this is just stale.

PR #1288, opened 2016-06-09
[Fresco][RFC] Fix rounded path

plamenko:
This clips the rounded path to the actual image bounds (if smaller than view) so that we don't get repeated edges. Unfortunately, it only works on API 19.

### DIFF
--- a/drawee/src/main/java/com/facebook/drawee/drawable/RoundedBitmapDrawable.java
+++ b/drawee/src/main/java/com/facebook/drawee/drawable/RoundedBitmapDrawable.java
@@ -23,9 +23,12 @@ import android.graphics.ColorFilter;
 import android.graphics.Matrix;
 import android.graphics.Paint;
 import android.graphics.Path;
+import android.graphics.Rect;
 import android.graphics.RectF;
+import android.graphics.Region;
 import android.graphics.Shader;
 import android.graphics.drawable.BitmapDrawable;
+import android.os.Build;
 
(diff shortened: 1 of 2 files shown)


drawee/src/main/java/com/facebook/drawee/drawable/RoundedBitmapDrawable.java
```diff
@@ -283,6 +286,10 @@ private void updateTransform() {
 
   private void updatePath() {
     if (mIsPathDirty) {
+      mBitmapBounds.set(0, 0, getBitmap().getWidth(), getBitmap().getHeight());
+      mTransform.mapRect(mBitmapBounds);
+      //mRootBounds.intersect(mBitmapBounds);
```
plamenko: This line should be removed. Uncommenting it would make the rounded path be like on the picture, but this is actually not desirable because of the border. Namely, there might be other branches such as background or placeholder and all of them should have the same rounded border.
 
screen shot 2016-06-08 at 20 02 40

plamenko:
This is how it looks after the patch:

screen shot 2016-06-08 at 20 04 04

plamenko:
Another solution for repeated edges is to intersect bitmap rect with drawable bounds before making a rounded path (like in the commeted out line), but then only round the corners that were not clipped. This becomes problematic if the clipping doesn't completely cover the corners.

kirwan:
I'm closing our very old PRs from internal people. We may still pursue this but right now this is just stale.

PR #1328, opened 2016-06-28
[RFC] RetainingDataSourceSupplier

plamenko:
`RetainingDataSourceSupplier` retains the previous image until the new image is ready. This is useful in slideshow and similar use cases.

### DIFF
--- a/fbcore/src/main/java/com/facebook/datasource/RetainingDataSourceSupplier.java
+++ b/fbcore/src/main/java/com/facebook/datasource/RetainingDataSourceSupplier.java
@@ -1,0 +1,140 @@
+package com.facebook.datasource;
+
+import com.facebook.common.executors.CallerThreadExecutor;
+import com.facebook.common.internal.Supplier;
+
+import java.util.Collections;
+import java.util.Set;
+import java.util.WeakHashMap;
+
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
+import javax.annotation.concurrent.NotThreadSafe;
+import javax.annotation.concurrent.ThreadSafe;
+
+@NotThreadSafe
+public class RetainingDataSourceSupplier<T> implements Supplier<DataSource<T>> {
+
+  private final Set<RetainingDataSource> mDataSources =
+      Collections.newSetFromMap(new WeakHashMap<RetainingDataSource, Boolean>());
+
+  private Supplier<DataSource<T>> mCurrentDataSourceSupplier = null;
+
+  @Override
+  public DataSource<T> get() {
+    RetainingDataSource dataSource = new RetainingDataSource();
+    dataSource.setSupplier(mCurrentDataSourceSupplier);
+    mDataSources.add(dataSource);
+    return dataSource;
+  }
+
+  public void setSupplier(Supplier<DataSource<T>> supplier) {
+    mCurrentDataSourceSupplier = supplier;
+    for (RetainingDataSource dataSource: mDataSources) {
+      dataSource.setSupplier(supplier);
+    }
+  }
+
+  @ThreadSafe
+  private class RetainingDataSource extends AbstractDataSource<T> {
+    @GuardedBy("RetainingDataSource.this")
+    @Nullable
+    private DataSource<T> mDataSource = null;
+
+    public void setSupplier(@Nullable Supplier<DataSource<T>> supplier) {
+      // early return without calling {@code supplier.get()} in case we are closed
+      if (isClosed()) {
+        return;
+      }
+      DataSource<T> oldDataSource;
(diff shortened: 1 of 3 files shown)


fbcore/src/main/java/com/facebook/datasource/RetainingDataSourceSupplier.java
```diff
@@ -0,0 +1,140 @@
...
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.GuardedBy;
+import javax.annotation.concurrent.NotThreadSafe;
+import javax.annotation.concurrent.ThreadSafe;
+
+@NotThreadSafe
+public class RetainingDataSourceSupplier<T> implements Supplier<DataSource<T>> {
+
+  private final Set<RetainingDataSource> mDataSources =
+      Collections.newSetFromMap(new WeakHashMap<RetainingDataSource, Boolean>());
```
lambdapioneer: We should be fine just using an `ArrayList` here

fbcore/src/main/java/com/facebook/datasource/RetainingDataSourceSupplier.java
```diff
@@ -0,0 +1,140 @@
...
+
+  @Override
+  public DataSource<T> get() {
+    RetainingDataSource dataSource = new RetainingDataSource();
+    dataSource.setSupplier(mCurrentDataSourceSupplier);
+    mDataSources.add(dataSource);
+    return dataSource;
+  }
+
+  public void setSupplier(Supplier<DataSource<T>> supplier) {
```
lambdapioneer: Perhaps name this `replaceSupplier(...)` to make it clear that this method is meant to be called several times?

fbcore/src/main/java/com/facebook/datasource/RetainingDataSourceSupplier.java
```diff
@@ -0,0 +1,140 @@
...
+
+    public void setSupplier(@Nullable Supplier<DataSource<T>> supplier) {
+      // early return without calling {@code supplier.get()} in case we are closed
+      if (isClosed()) {
+        return;
+      }
+      DataSource<T> oldDataSource;
+      DataSource<T> newDataSource = (supplier != null) ? supplier.get() : null;
+      synchronized (RetainingDataSource.this) {
+        if (isClosed()) {
```
lambdapioneer: Can we rewrite this s.t. some of the side-effects (e.g. `oldDataSource = newDataSource` will make `newDataSource` being closed below) clearer? For example with an early return:

```java
if (isClosed()) {
  closeSafely(newDataSource);
  return;
}

File: fbcore/src/main/java/com/facebook/datasource/RetainingDataSourceSupplier.java [hunk 4]
```diff
@@ -0,0 +1,140 @@
...
+
+    @Override
+    public synchronized boolean hasResult() {
+      return (mDataSource != null) && mDataSource.hasResult();
+    }
+
+    @Override
+    public boolean close() {
+      DataSource<T> dataSource;
+      synchronized (RetainingDataSource.this) {
```
lambdapioneer: `this` is sufficient here (same applies above in line 51)

lambdapioneer:
Looks cool and will be definitely helpful for the use-cases where we face the `flashing` images issues. I didn't comment on the change to the sample app, as I guess that you'll update that part anyway.

chrisjenx:
What's the status of this? Also when using this currently with `.setControllerListener(object : BaseControllerListener<ImageInfo>()` you don't get callbacks.

athornz:
Any progress on this? Would love to see this merged.

oprisnik:
For now, you can just copy paste `RetainingDataSourceSupplier.java` from the PR, should just work. We'll take a look again at this when we have some time.

stellio-team:
Would be good to have it.

zmroczek:
This was closed in the commit [ff3bed9](https://github.com/facebook/fresco/commit/ff3bed97fb05e4d235e5e67009b1e26b7ad384b8).

Crysis21:
I'm using fresco to display gifs in a `RecyclerView`. After the user has tapped an imaged I open a details screen where I should display in the end a high-resolution GIF.

If I just load another GIF in the same `DraweeView` the content will disappear as this is the way Fresco works and the user will see a black image while the high res one loads, although I already have a low quality preview available.

I also tried using the low-res/high-res scheme, but while the high res image is loading, the low-res is not animated.

Later I discovered the `RetainingDataSourceSupplier` in Fresco, which works by keeping the current image in a `DraweeView` until a new one is loaded. This works fine but it doesn't seem to work for animated content.

Can you help me retain the current animated content until the high quality one is downloaded?

Issue #1317, opened 2016-06-21
[question] determineDownsampleRatio is inappropriate 


huangjunkun:
> reference：com.facebook.imagepipeline.producers.DownsampleUtil#determineDownsampleRatio

I think that determineDownsampleRatio is inappropriate.

`float ratio = Math.max(widthRatio, heightRatio);`

It's could cause ImagePipeline use more Memory to decode image for bitmap, a wasteful behavior.
I think the logical thing to do would have been to select which come near to 1.0f, eg. 0.6 vs 1.2 and select 0.9.

huangjunkun:
`float ratio = (Math.abs(1.0f-widthRatio) > Math.abs(1.0f-heightRatio)) ? widthRatio : heightRatio;`

If you consider that is reasonable, **I can make a pull request**.

oprisnik:
The DownsampleUtil is used by the ImagePipeline itself, which does not know anything on how the image is used later on.
I suppose that for your case, you're using a `fit...` scaling type for your DraweeView? 
But the pipeline at this point does not know that (and different DraweeViews could have different scale types) and if we're too aggressive with downscaling and you're using cropping, the quality might be worse.

huangjunkun:
Sorry. You maybe not understand my content.

determineDownsampleRatio return ration, which is for adjust sample. Look the following
> com.facebook.imagepipeline.producers.DownsampleUtil#determineSampleSize

`  float ratio = determineDownsampleRatio(imageRequest, encodedImage);
    int sampleSize;
    if (encodedImage.getImageFormat() == ImageFormat.JPEG) {
      sampleSize = ratioToSampleSizeJPEG(ratio);
    } else {
      sampleSize = ratioToSampleSize(ratio);
    }`

and ratio > 0.8333, sampleSize = 1.//decode a original Bitmap.
> com.facebook.imagepipeline.producers.DownsampleUtil#ratioToSampleSize
`  if (ratio > 0.5f + 0.5f * INTERVAL_ROUNDING) {
      return 1; // should have resized
    }`

huangjunkun:
Example

>  ratio 0.6 vs 1.2 and  a bitmap of 1920*1080 (used memory 1920*1080*3  approximately equal to 6Mb)

`float ratio = Math.max(widthRatio, heightRatio);`
select ratio 1.2 and sampleSize=1, used memory 6Mb (approximately)


[...]

PR #1330, opened 2016-06-30
[OPT] optimize determineDownsampleRatio for saving used memory

huangjunkun:
optimize determineDownsampleRatio for saving used memory.
issues:https://github.com/facebook/fresco/issues/1317


kirwan:
We have a plan for a slightly more sophisticated way to control resizing. With that, I don't think this change is necessary.

Issue #1091, opened 2016-03-29
BUG for mDecodeFileDescriptorEnabledForKitKat


huangjunkun:
LocalFetchProducer
    if (mDecodeFileDescriptorEnabledForKitKat && inputStream instanceof FileInputStream &&
        javaMax >= 64 * javaFree) {
      return getInputStreamBackedEncodedImage(new File(inputStream.toString()), length);
    }

result in ：
com.facebook.imagepipeline.platform.DalvikPurgeableDecoder#decodeJPEGFromEncodedImage

    final CloseableReference<PooledByteBuffer> bytesRef = encodedImage.getByteBufferRef();
    Preconditions.checkNotNull(bytesRef);

throw new NullPointerException();
==========================
so：
 mDecodeFileDescriptorEnabledForKitKat is faulted commit ！  ：（

aagnes:
THanks for pointing it out. We'll fix the bug asap.

huangjunkun:
OK

kirwan:
This is actually an experimental config option which I wouldn't currently recommend using.

In the next release this has been moved to a separate class to the main `ImagePipelineConfig` called [`ImagePipelineExperiments`](https://github.com/facebook/fresco/blob/abeb080ce8135326a450e65b2e00b95f9dacd876/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipelineExperiments.java) to make it more obvious that this is one of a few options which are riskier to change.

huangjunkun:
A good idea.

huangjunkun:
Hoo ~ nice.

PR #1331, opened 2016-06-30
[OPT]optimize LocalFetchProducer & LocalFileFetchProducer

huangjunkun:
[OPT]optimize LocalFetchProducer & LocalFileFetchProducer

> isMemoryEnough
> 
> https://github.com/facebook/fresco/issues/1091


imagepipeline/src/main/java/com/facebook/imagepipeline/producers/LocalFetchProducer.java
```diff
@@ -105,20 +105,34 @@ protected EncodedImage getByteBufferBackedEncodedImage(
     }
   }
 
-  protected EncodedImage getEncodedImage(
+    private boolean isMemoryEnough() {
+      Runtime runTime = Runtime.getRuntime();
+      long javaMax = runTime.maxMemory();
+      long javaUsed = runTime.totalMemory() - runTime.freeMemory();
+      long javaFree = Math.min(javaMax - javaUsed, 8 * ByteConstants.MB);
+      return (javaMax >= 64 * javaFree);
```
duchuyctlk: Based on the method name, I can interfere that if the free memory is less than or equal to the maximum memory, then there is enough memory. It doesn't sound very logical.

huangjunkun: @duchuyctlk   **javaMax** contain more free java free memory in the future. **javaFree**  is not constant.
 ` (javaMax >= 64 * javaFree)`  means that java heap has enough memory.

kirwan:
The code this was changing has mostly now been removed.

PR #1385, opened 2016-07-28
Added support for synchronous small disk cache check

KoMinkyu:
Added a support to check in small disk cache synchronously.
Without this change, there is no way to check cache in small disk cache.
So I added this support.

### DIFF
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipeline.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipeline.java
@@ -454,10 +454,13 @@ public class ImagePipeline {
     final CacheKey cacheKey = mCacheKeyFactory.getEncodedCacheKey(imageRequest, null);
     final ImageRequest.CacheChoice cacheChoice = imageRequest.getCacheChoice();
 
-    if (cacheChoice == ImageRequest.CacheChoice.SMALL) {
-      return mSmallImageBufferedDiskCache.diskCheckSync(cacheKey);
-    } else {
-      return mMainBufferedDiskCache.diskCheckSync(cacheKey);
+    switch (cacheChoice) {
+      case DEFAULT:
+        return mMainBufferedDiskCache.diskCheckSync(cacheKey);
+      case SMALL:
+        return mSmallImageBufferedDiskCache.diskCheckSync(cacheKey);
+      default:
+        return false;
     }
   }
 
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipeline.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipeline.java
@@ -44,7 +44,6 @@ import com.facebook.imagepipeline.request.ImageRequestBuilder;
 
 import bolts.Continuation;
 import bolts.Task;
-
 import com.android.internal.util.Predicate;
 
 /**
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipeline.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipeline.java
@@ -423,7 +423,8 @@ public class ImagePipeline {
    * @return true if the image was found in the disk cache, false otherwise.
    */
   public boolean isInDiskCacheSync(final Uri uri) {
-    return isInDiskCacheSync(uri, ImageRequest.CacheChoice.DEFAULT);
+    return isInDiskCacheSync(uri, ImageRequest.CacheChoice.SMALL) ||
+            isInDiskCacheSync(uri, ImageRequest.CacheChoice.DEFAULT);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


facebook-github-bot-5:

facebook-github-bot-5:
@KoMinkyu updated the pull request.

imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipeline.java
```diff
@@ -423,7 +423,24 @@ public boolean isInBitmapMemoryCache(final ImageRequest imageRequest) {
    * @return true if the image was found in the disk cache, false otherwise.
    */
   public boolean isInDiskCacheSync(final Uri uri) {
-    return isInDiskCacheSync(ImageRequest.fromUri(uri));
+    return isInDiskCacheSync(uri, ImageRequest.CacheChoice.DEFAULT);
```
lambdapioneer: Can we check both caches here? `isInDiskCacheSync(uri, CacheChoice.SMALL) || isInDiskCacheSync(uri, CacheChoice.DEFAULT)`

As we have the in-memory index for the disk cache now, this should be much faster anyway

imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipeline.java
```diff
@@ -435,7 +452,13 @@ public boolean isInDiskCacheSync(final Uri uri) {
    */
   public boolean isInDiskCacheSync(final ImageRequest imageRequest) {
     final CacheKey cacheKey = mCacheKeyFactory.getEncodedCacheKey(imageRequest, null);
-    return mMainBufferedDiskCache.diskCheckSync(cacheKey);
+    final ImageRequest.CacheChoice cacheChoice = imageRequest.getCacheChoice();
+
+    if (cacheChoice == ImageRequest.CacheChoice.SMALL) {
```
lambdapioneer: style: Let's make this a `switch`. If we ever happen to add something to `CacheChoice`, the compiler will warn us if we forget about this location.

lambdapioneer:
Thanks @KoMinkyu for the PR :) I've added some comments inline. To make this PR awesome, please also update/extend the current test cases in `BufferedDiskCacheTest`: https://github.com/facebook/fresco/blob/a80a7ec2e68c7e21531d9d49b0eb042a04e52b20/imagepipeline/src/test/java/com/facebook/imagepipeline/cache/BufferedDiskCacheTest.java#L119-L129

KoMinkyu:
Hello @lambdapioneer. Thank you for your comments.
I will make change and update :)

imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipeline.java
KoMinkyu: If I followed code correctly, `isInDiskCacheSync` method finally check disk cache with file check.

If you see `diskCheckSync` method with this [link](https://github.com/facebook/fresco/blob/master/imagepipeline/src/main/java/com/facebook/imagepipeline/cache/BufferedDiskCache.java#L113-L123) in `BufferedDiskCache`, that code is double checking in-memory index and file cache.

Because of calling ` checkInStagingAreaAndFileCache(CacheKey)`, it will check file cache by [this line](https://github.com/facebook/fresco/blob/master/imagepipeline/src/main/java/com/facebook/imagepipeline/cache/BufferedDiskCache.java#L158) if there is no index in in-memory index list.

So I thought, if there are someone who using this method currently, this change would make different result and more slow performance. So I just make it to check `DEFAULT`  disk cache if no cache-type provided.

Is there any problem with that?

KoMinkyu:
Hi @lambdapioneer ! Thank you for your review.
I commented some opinion under your comment. If you think there is no matter with that, I will update my PR with what you suggested :)

thanks!

facebook-github-bot-6:
@KoMinkyu updated the pull request.

lambdapioneer:
Thanks a lot for the update!

I'd argue that `isInDiskCacheSync(final Uri uri)` is currently giving a wrong/bad answer. I'd expect that people using this method usually want to know if the image is in any disk cache s.t. they can route some application logic. An example would be to show articles first where images are available w/o network fetch. Currently, their code might make a non-optimal decision.

The async `DataSource<Boolean> isInDiskCache(final Uri uri)` also checks both caches.

What I meant with the "in memory cache" is that we often don't need to do a full disk-sweep for the cache check duet to this recent improvement: https://github.com/facebook/fresco/commit/c5e8a450527f3566de48198b836e66df6661847e

This method is executed before the slow real check is performed: https://github.com/facebook/fresco/blob/4e52672dcde4821254f870935dcf0d49156bd626/imagepipeline-base/src/main/java/com/facebook/cache/disk/DiskStorageCache.java#L582-L595

KoMinkyu:
Ah..! I think I've misunderstood. Now I could get good understanding about it through your kind explanation. 
I will update my PR ASAP. thanks :)

facebook-github-bot-9:
@KoMinkyu updated the pull request.

lambdapioneer:
Awesome! I'll import it and maybe adapt some parts to our style guide. You'll see an update here once we've merged it

@facebook-github-bot import

facebook-github-bot-8:
Thanks for importing.

Issue #1176, opened 2016-04-30
Fresco crashing on using SVG as placeHolder and failure images prior to lollipop


embydextrous:
Since I upgraded to support library 23.3.0. I am using only vector drawables and not png assets. Working fine on lollipop and above but crashing prior to lollipop.

oprisnik:
Hi! 
We currently do not support vector drawables for API < 21. 
It works for API 21 and above because Android added native vector drawable support with Lollipop. The support library internally replaces your `ImageView` with `AppCompatImageView` (which supports vectors pre-Lollipop) but it can't do this for `DraweeView`. 
You can also take a look at https://medium.com/@chrisbanes/appcompat-v23-2-age-of-the-vectors-91cbafa87c88#.hwhg5l1rd for more information.

embydextrous:
Okay. Any plans to do this in near future.
I am a huge fan of fresco image loading and want to make use of xml resources to reduce app size.
I don't want to move to any other library.

oprisnik:
A bit of a clarification there: By "We currently do not support vector drawables for API < 21." I meant we do not support vector drawables if specified **in XML**.
You can of course still us Fresco with vector drawables on all platform by building your own hierarchy.

A quick example for this would be:

```java
// android.support.v4.content.ContextCompat creates your vector drawable
Drawable placeholderDrawable = ContextCompat.getDrawable(context, R.drawable.yourVectorAsset);

SimpleDraweeView draweeView = (SimpleDraweeView) findViewById(R.id.yourImage);

// Set the placeholder image to the placeholder vector drawable
draweeView.setHierarchy(
    GenericDraweeHierarchyBuilder.newInstance(context.getResources())
        .setPlaceholderImage(placeholderDrawable)
        .build());
```

oprisnik:
I just tested this and it also works for me if I define the placeholder in XML

```xml
<com.facebook.drawee.view.SimpleDraweeView
    android:id="@+id/drawee"
    ...
    fresco:placeholderImage="@drawable/vector_graphic"
    />
```
with

`compile 'com.android.support:appcompat-v7:23.2.1'`

and 
```
android {
    defaultConfig { 
        vectorDrawables.useSupportLibrary = true  
    }
}
```
in my build.gradle.

embydextrous:
Thank you so much. This worked for me.

embydextrous:
Is it working before lollipop?

oprisnik:
[...]

Issue #1008, opened 2016-02-21
Why don't support SVG 


johnzheng00:
????  Why don't support SVG ???

aagnes:
There's little use request for SVG so far. You may be the only one asking for this.

johnzheng00:
Yes, android just released new support v7 library.
 Android Support Library, revision 23.2.0 (February 2016)
http://developer.android.com/tools/support-library/index.html
Adds support for VectorDrawable assets to apps running on API level 7 or higher. AnimatedVectorDrawable assets are also supported on API level 11 or higher. Vector assets can be considerably smaller than image assets and should help reduce app size by reducing the amount of assets required to support multiple device screens.

Mauker1:
He's definitely not the only one looking for SVG support. It would be nice if Fresco offered support for it. I had to migrate to Glide because I needed this feature.

KynoYang:
I use a animated vector drawable as the progressBarImage since vector drawable has been support by the android support library, however, it seems that fresco doesn't support to use vector.

scottkruse:
Please put this in the next sprint, much needed!

johnzheng00:
Android Studio have a tool, New->Vector Asset->      change the svg to vectordrawable, use appcompat v7  23.2.0 up, app:srcCompat="@drawable/vectordrawable" . So we can used svg to vectordrawable
Android Support Library, revision 23.2.0 (February 2016)
http://developer.android.com/tools/support-library/index.html

aagnes:
We currently have no plan in support SVG at the moment. We would appreciate if someone in community wants to contribute it.

oprisnik:
You can use vector drawables from the support library. See #1176 for more information

chethann:
As fresco doesn't have support for SVGs, vector drawables can't be used with React native as well. Had to convert many svgs into pngs in order to work with React native.

oprisnik:
We're looking into adding custom decoders for Fresco so that you can add an SVG decoder if needed.

oprisnik:
You can now add custom SVG decoders and drawable factories now. A very simple example how this could look like can be found here: https://github.com/facebook/fresco/tree/master/samples/showcase/src/main/java/com/facebook/fresco/samples/showcase/imageformat/svg

chandanrana:
[...]

Issue #1000, opened 2016-02-17
Images are not appearing back in list view when switching between tabs


madkid66:
We just started using Fresco in our app. The UI has different tabs with list views in each with an image (SimpleDrawee) in each cell. The initial load happens fine. But, when we switch between tabs and go to another screen and come back, some of the SimpleDrawees are not showing up on the screen. Please reply with any suggestions at the earliest. I am attaching the screenshot and the logs (verbose - as suggested on the Troubleshooting section of frescolib.org.
Thank you
![screenshot_2016-02-17-10-35-14](https://cloud.githubusercontent.com/assets/10780191/13123476/1fd1a1e4-d571-11e5-99b9-278cb1bb7ea4.png)
[logs.txt](https://github.com/facebook/fresco/files/135084/logs.txt)

IanChilds:
Does the image ever load? Is there just a delay or does it stay black? Also, if you scroll up and down does the image become visible?

madkid66:
Thank you IanChilds for the reply. I noticed that the images load first time (although few of them take time   ). But, the issue is when I switch between tabs or go to another screen (clicking the images takes to another screen) and return to the list, the images are blank. When I scroll up and down, the images appear back most of the time (although not always).

madkid66:
IanChilds,

Please help reply with any suggestions. We are seeing more of it. 
Thank you

![screenshot_2016-02-19-18-30-09](https://cloud.githubusercontent.com/assets/10780191/13193963/581e53fe-d737-11e5-9f31-7652a319243f.png)

IanChilds:
From your logs, it looks like the only errors are coming from `ChatterAndTweetsAdapter` - could that be relevant? 

Perhaps your cache is getting full? Are you holding onto a lot of images in other tabs that are not on the screen? Can you try increasing the size of the cache?

madkid66:
The messages related to ChatterAndTweetsAdapter are not related.
From the logs I see that certain images are going through 'clearImage' when not visible, but are not appearing back (and any associated log) when they are supposed to be visible again.

michalgr:
I wonder if all the drawee views get notification of being reattached. It might be worth looking at part of the logcat that correspond to "going back" to the affected tab. Could you please do this to get us more debugging data:
1. open any tab in your app and wait for all the images to show up
2. go to another tab and wait for all the images to show up
3. clear logcat (`adb logcat -c`)
4. start recording logcat
5. go back to the first tab
6. check that some images do not show up
[...]

Issue #329, opened 2015-05-28
should support svg and vector  xml function.


johnzheng00:
the fresco should support svg and vector xml function!?

TarangS:
@johnzheng00 it would be useful if you could tell us the use case which fresco falls short on.

tyronen:
Fundamentally Fresco is a raster graphics library, not a vector images library. It's not likely we'd ever support this.

johnzheng00:
If use svg or vector, I just one vector xml or svg file for my project. But now we use png for my project. Support so many devices and screen size, mdpi to xxxdpi, over 4 size at least. So the png need 4 at least. My apk will be large.  Is it right ?

PR #1386, opened 2016-07-29
Svg

Roma-Lis:
I'd like to add svg rendering functionality to fresco.
I believe it should be useful because of issues ([#329](https://github.com/facebook/fresco/issues/329), [#1000](https://github.com/facebook/fresco/issues/1008), [#1176](https://github.com/facebook/fresco/issues/1176)).
The svg renderer implementation taken from [androidsvg project](http://bigbadaboom.github.io/androidsvg)


oprisnik:
Thanks for the PR @Roma-Lis!
We're looking into adding custom decoders for Fresco so that you can simply plug in your own decoders, like this one for SVG.
The format shouldn't be too different from the current implementation, so it should be pretty easy to adapt your code once this is ready.

We'll keep you updated!

Roma-Lis:
@oprisnik , thanks. Actual architecture don't support it, but it will be nice to know about view dimensions inside any vector images decoder to avoid redundant rendering.

oprisnik:
Yeah, we are working on making pluggable decoders. We don't know the view dimensions before, but we can enforce that resize options have to be set for the `ImageRequest`.

Roma-Lis:
@oprisnik sure, if it is not a problem to have third-party libraries added via Gradle dependencies in your project.

kirwan:
We now have custom decoders and will be adding a sample very soon for using SVG within Fresco.

I'm closing this pull request but please take a look later in the week.

oprisnik:
A very simple example for SVG support (using the same library as you've used in this PR) can be found here: https://github.com/facebook/fresco/tree/master/samples/showcase/src/main/java/com/facebook/fresco/samples/showcase/imageformat/svg 

I didn't check if the functionality is exactly the same as in your `SvgDecoder`, but you can just modify the custom decoder to achieve whatever you want.

Thanks for your time!

PR #1400, opened 2016-08-04
Use initial capaticy constructor when it's appropriate

pungrue26:
When you create a <code>ArrayList</code> or <code>HashSet</code>, you have to use constructor with
initial capacity argument if you know its final size beforehand.
Because we can avoid enlarging its backing array. This becomes critical
when the size of list becomes larger. For example, the <code>mBitmaps</code>
field in <code>CloseableAnimatedBitmap</code> class can hold many bitmaps, because
it holds bitmaps of animation.

Many classes in this project already follow such usage. I found
some classes that do not follow it, and changed those.


oprisnik:
Thank you very much @pungrue26, looks good!

@facebook-github-bot import

facebook-github-bot-5:
Thanks for importing.

pungrue26:
@oprisnik Nice to hear that! 

BTW, I just applied to [Software Engineer - Android position in London office](https://www.facebook.com/careers/jobs/a0IA0000006cPE9MAM/). I hope I can get an interview. :)

PR #1429, opened 2016-08-23
added original dimensions to ImageInfo

dbnoise:
The idea of pull-request is to expose original dimensions parsed by EncodedImage to ImageInfo, so UI level code can have access to this data.

### DIFF
--- a/animated-base/src/main/java/com/facebook/imagepipeline/image/CloseableAnimatedBitmap.java
+++ b/animated-base/src/main/java/com/facebook/imagepipeline/image/CloseableAnimatedBitmap.java
@@ -160,4 +160,19 @@ public class CloseableAnimatedBitmap extends CloseableBitmap {
     return (bitmaps == null) ? 0 : bitmaps.get(0).getHeight();
   }
 
+  /**
+   * @return width of original (non-resized) image
+   */
+  @Override
+  public int getOriginalWidth() {
+      return getWidth();
+  }
+
+  /**
+   * @return height of original (non-resized) image
+   */
+  @Override
+  public int getOriginalHeight() {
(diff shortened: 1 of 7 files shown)


facebook-github-bot-3:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

facebook-github-bot-9:

oprisnik:
Thanks for the pull request @dbnoise. Could you tell me why you would need to know the original image size?

jmbusby:
I have a use case for such a change. We have an image upload process on the client, that needs to view the image on device, and upload it. For viewing, the image is downsampled, but the original full size image is uploaded. If Fresco could return the original image dimensions when viewing the downsampled image, I could use that information to determine if the image is too large to upload, and get rid of the code I have that decodes the bounds only of the image in storage to get that size info.

kirwan:
I've just been reviewing this PR again and I can see how this could be useful.

However, Fresco supports two approaches to resizing images and this PR only covers the downsampling approach. The alternative involves transcoding to a smaller encoded image in the `ResizeAndRotateProducer` and if that happens the "original" width and height returned would be the smaller values from the `EncodedImage` created there.

As things stand I don't see a simple way to pass those original values but I'd welcome your thoughts if you think this pull request could be amended to cover that case too.

kirwan:
This has been marked as needs-revision for over six months so I'm going to close. If someone still has a strong need for something like this, please make the necessary changes and re-open.

PR #1432, opened 2016-08-25
Reduce unnecessary redraws when the ProgressBarDrawable is not set

### LINKED ISSUE #1417: FadeDrawable call invalidateSelf() so many times when the picture is downloading.
I had a RecyclerView like this.
screen_shot

I write a custom view follow this [guide](http://frescolib.org/docs/writing-custom-views.html) to show 0-9 pictures.
To make the view scroll smooth, I set the fade animation's duration is 0 and not set the ProgressBarDrawable. 

I except 
the custom view call onDraw() times = 1(init) + DraweeHolder count * (1(placeholder drawable) + 1(actual drawable))

But I find the invalidateDrawable() method called so many times.

I checked the source code(GenericDraweeHierarchy.java line 278)
```
  public void setProgress(float progress, boolean immediate) {
    mFadeDrawable.beginBatchMode();
    setProgress(progress);
    if (immediate) {
      mFadeDrawable.finishTransitionImmediately();
    }
    mFadeDrawable.endBatchMode();
  }

```
The InvalidateSelf() function called wether the ProgressBarDrawable is set or not.
I think we can check ProgressBarDrawable not null first, then call the ```setProgress ``` function.

s1rius:
Fix #1417

### DIFF
--- a/drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
+++ b/drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
@@ -224,7 +224,7 @@ public class GenericDraweeHierarchy implements SettableDraweeHierarchy {
   }
 
   private void setProgress(float progress) {
-    Drawable progressBarDrawable = getParentDrawableAtIndex(PROGRESS_BAR_IMAGE_INDEX).getDrawable();
+    Drawable progressBarDrawable = mFadeDrawable.getDrawable(PROGRESS_BAR_IMAGE_INDEX);
     if (progressBarDrawable == null) {
       return;
     }
@@ -275,6 +275,9 @@ public class GenericDraweeHierarchy implements SettableDraweeHierarchy {
 
   @Override
   public void setProgress(float progress, boolean immediate) {
+    if (mFadeDrawable.getDrawable(PROGRESS_BAR_IMAGE_INDEX) == null) {
+      return;
+    }
     mFadeDrawable.beginBatchMode();
     setProgress(progress);
     if (immediate) {
@@ -425,7 +428,7 @@ public class GenericDraweeHierarchy implements SettableDraweeHierarchy {
    * @return true if there is a placeholder image set.
    */
   public boolean hasPlaceholderImage() {
-    return getParentDrawableAtIndex(PLACEHOLDER_IMAGE_INDEX) != null;
+    return mFadeDrawable.getDrawable(PLACEHOLDER_IMAGE_INDEX) != null;
   }
 
   /** Sets the placeholder image focus point. */


oprisnik:
Thanks @s1rius , looks good!

oprisnik:
@facebook-github-bot import

facebook-github-bot-6:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3770600) to review internal test results.

kirwan:
This does look good.

However, I've looked at `getParentDrawableAtIndex()` and realised that will create an object just to then check if the thing inside is null. A `hasDrawableAtIndex()` method would be much better.

facebook-github-bot-8:
@s1rius updated the pull request - [view changes](https://github.com/facebook/fresco/pull/1432/files/78361a4091a31d74cbc7421e0afbdb9f3c06cecd..a578a752286671a591b27d5a7aa4d35b73e89aab) - [changes since last import](https://github.com/facebook/fresco/pull/1432/files/78361a4091a31d74cbc7421e0afbdb9f3c06cecd..a578a752286671a591b27d5a7aa4d35b73e89aab)

s1rius:
Hi @oprisnik @kirwan , thanks for review. I updated the PR.

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
```diff
@@ -334,6 +336,13 @@ private DrawableParent getParentDrawableAtIndex(int index) {
   }
 
   /**
+   * Returns whether the given layer has a child drawable.
+   */
+  private boolean hasChildDrawableAtIndex(int index) {
+    return getParentDrawableAtIndex(index).getDrawable() != null;
```
kirwan: @s1rius Thanks for the change but it doesn't address what I noticed.

This calls through to `ArrayDrawable.getDrawableParentForIndex(int)` which will create the `ParentDrawable` against which you then extract the null child.

I think there are two better options:

1) add a method to `ArrayDrawable` called something like `hasDrawableAtIndex(int)` which just checks `mLayers[index]` for null and call that against `mFadeDrawable`; or
2) call `mFadeDrawable.getDrawable(int)` (as called for example in `setFailure(Throwable)`) which is equivalent to what you're doing here without have any side-effects

I realise that this approach existed previously in `hasPlaceholderImage()` but it would be great to improve on that while making this change. For bonus points, it looks like `setProgress(float)` could make use of this new method too.

facebook-github-bot-7:
@s1rius updated the pull request - [view changes](https://github.com/facebook/fresco/pull/1432/files/a578a752286671a591b27d5a7aa4d35b73e89aab..6c0b1149ab47853c483525663804ae0dd12ea938) - [changes since last import](https://github.com/facebook/fresco/pull/1432/files/a578a752286671a591b27d5a7aa4d35b73e89aab..6c0b1149ab47853c483525663804ae0dd12ea938)

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
```diff
@@ -421,13 +423,6 @@ public void setPlaceholderImage(Drawable drawable, ScaleType scaleType) {
     getScaleTypeDrawableAtIndex(PLACEHOLDER_IMAGE_INDEX).setScaleType(scaleType);
   }
 
-  /**
-   * @return true if there is a placeholder image set.
-   */
-  public boolean hasPlaceholderImage() {
-    return getParentDrawableAtIndex(PLACEHOLDER_IMAGE_INDEX) != null;
-  }
-
```
kirwan: Please just bring this method back to avoid breaking the public API in case anyone is using it. Then we can merge this.

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
```diff
@@ -275,12 +275,14 @@ public void setImage(Drawable drawable, float progress, boolean immediate) {
 
   @Override
   public void setProgress(float progress, boolean immediate) {
-    mFadeDrawable.beginBatchMode();
-    setProgress(progress);
-    if (immediate) {
-      mFadeDrawable.finishTransitionImmediately();
+    if (mFadeDrawable.getDrawable(PROGRESS_BAR_IMAGE_INDEX) != null) {
```
kirwan: This is purely a nit but it might be best to do the same as in `setProgress(float)` where it does an early return if the drawable is null rather than surrounding the remainder of the method in the if block.

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
s1rius: No one used it before. And this method always return true whether the place holder image is set or not.

kirwan: I meant that an app using Fresco might be calling the method. In fact we do have a caller to this in our codebase.

You're right that it doesn't behave as expected but I'll deal with that.

s1rius: Oh, Could I change it to 
public boolean hasPlaceholderImage() {
  return mFadeDrawable.getDrawable(PLACEHOLDER_IMAGE_INDEX) != null;
}

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
s1rius: Sorry for my poor english, I dont understand.

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
kirwan: That would be perfect.

drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchy.java
kirwan: I meant to do this...

```
if (mFadeDrawable.getDrawable(PROGRESS_BAR_IMAGE_INDEX) == null) {
  return;
}

// Rest of method as before
```

s1rius: OK

kirwan:
Thanks for your patience with me. This looks great now.

@facebook-github-bot import

facebook-github-bot-3:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3770600) to review internal test results.

s1rius:
Thanks for your patience, too.

PR #1443, opened 2016-08-31
Update proguard-fresco.pro

PiotrWpl:
I believe, we don't need it anymore.


kirwan:
You're right, but I'd prefer to leave it for a little longer in case people are still using the older versions.

I'll merge this after completing the next release.

huangjunkun:
@PiotrWpl 
https://github.com/facebook/fresco/commit/d5f6b4266476db747b7a1497471c44f435f02016#commitcomment-18684613

Issue #1472, opened 2016-09-21
imagePipelineFactory.getMainFileCache().getSize() return -1 after app restart


eldk:
Hello,

Testing crash reports and then have some app restart, I see imagePipelineFactory.getMainFileCache().getSize() return -1 even if cache is not empty (it's the same for getCount()).

This was not the case some time ago.

My repository is up to https://github.com/facebook/fresco/commit/93d5a4c81c8b4b1bb9e5a1f874851969d06a3a0f.

I will update it this afternoon and test again.

Thanks,

Eric

eldk:
Hello,

I have tested on last Fresco master branch : https://github.com/facebook/fresco/commit/b74da9e78f47db5baaba419f2fea606607d6cb4b .

It's the same, after app restart (device restart, force shutdown, or app crash), cacheStats are not initialized, so getSize() or getCount() return -1.

This PR solve it : https://github.com/facebook/fresco/pull/1473.

Thanks,

Eric

PS : CircleCl is broken

lambdapioneer:
Hi @eldk,

I'm currently fixing the CI. Once that's done I'll have a look at the PR.

Cheers, Daniel

eldk:
Hello,

This was solved using : 

`DiskCacheConfig diskCacheConfig = DiskCacheConfig.newBuilder(context)`
(...)
`.setIndexPopulateAtStartupEnabled(true)`
`.build();`

Doc : http://frescolib.org/javadoc/reference/com/facebook/cache/disk/DiskCacheConfig.html

Sorry,

Eric

weibingyi:
Fresco.getImagePipelineFactory().getMainFileCache().trimToMinimum(); 
long cacheSizeFresco = Fresco.getImagePipelineFactory().getMainFileCache().getSize();

Issue #1460, opened 2016-09-07
DiskStorageCache mLock problem


jiqimaogou:
as we see, when new DiskStorageCache class, it could use a background thread to calculate the cache size which may take long time, and when I want to get a resource, it must wait the initialization to complete, because they use the same locker. 

can this be optimized?

erikandre:
Closing as part of our cleanup of old issue.
We've not seen that this synchronization (and waiting for the lock) is causing issues but if anyone is experiencing issue with this, please file a new bug report.

PR #1475, opened 2016-09-22
issue-1472:Add statement to check if mCacheStats isInitialized when getSize() or getCount()

eldk:
Hello,

#1472

Add statement to check if mCacheStats isInitialized in getSize() and getCount(), if not, call maybeUpdateFileCacheSize().

Thanks,

Eric

### DIFF
--- a/imagepipeline-base/src/main/java/com/facebook/cache/disk/DiskStorageCache.java
+++ b/imagepipeline-base/src/main/java/com/facebook/cache/disk/DiskStorageCache.java
@@ -592,10 +592,16 @@ public class DiskStorageCache implements FileCache, DiskTrimmable {
   }
 
   public long getSize() {
+    if (!mCacheStats.isInitialized()){
+      maybeUpdateFileCacheSize();
+    }
     return mCacheStats.getSize();
   }
 
   public long getCount() {
+    if (!mCacheStats.isInitialized()){
+      maybeUpdateFileCacheSize();
+    }
     return mCacheStats.getCount();
   }
 


lambdapioneer:
Hi @eldk,

thanks for the PR. Looks good 👍  Let me import it to run the tests as CI is broken at the current base of your diff.

facebook-github-bot-0:
Thanks for importing. If you are a Facebook employee, you can view this diff [on Phabricator](https://phabricator.intern.facebook.com/D3913333).

facebook-github-bot-4:
@eldk updated the pull request - [view changes](https://github.com/facebook/fresco/pull/1475/files/ad0732d073e65e404ae376448681b8d2102143b1..eb4451f9a97346e120df89aae6604efb4eb20e4e) - [changes since last import](https://github.com/facebook/fresco/pull/1475/files/ad0732d073e65e404ae376448681b8d2102143b1..eb4451f9a97346e120df89aae6604efb4eb20e4e)

eldk:
https://github.com/facebook/fresco/blob/master/imagepipeline-base/src/test/java/com/facebook/cache/disk/DiskStorageCacheTest.java#L298

imagepipeline-base/src/main/java/com/facebook/cache/disk/DiskStorageCache.java
```diff
@@ -592,10 +592,16 @@ private void updateFileCacheSizeLimit() {
   }
 
   public long getSize() {
+    if (!mCacheStats.isInitialized()){
+      maybeUpdateFileCacheSize();
```
lambdapioneer: We would have to have a `synchronized(mLock)` around this, as `maybeUpdateFileCacheSize()` is expected to be called in a synchronized manner (see line 703/709).

Alternatively, we could throw an `IllegalStateException` in those cases where the client has to wait for the initialization.

Which variant would you prefer? And what the use-cases would you anticipate where clients need to check this early on?

imagepipeline-base/src/main/java/com/facebook/cache/disk/DiskStorageCache.java
```diff
@@ -592,10 +592,16 @@ private void updateFileCacheSizeLimit() {
...
+    }
     return mCacheStats.getSize();
   }
 
   public long getCount() {
+    if (!mCacheStats.isInitialized()){
+      maybeUpdateFileCacheSize();
```
lambdapioneer: same here

eldk:
https://github.com/facebook/fresco/issues/1460

eldk:
Hello @lambdapioneer  ,

Setting DiskCacheConfig as this : 

`DiskCacheConfig diskCacheConfig = DiskCacheConfig.newBuilder(context)`
(...)
`.setIndexPopulateAtStartupEnabled(true)`
`.build();`

Doc : http://frescolib.org/javadoc/reference/com/facebook/cache/disk/DiskCacheConfig.html

Solve this : 
https://github.com/facebook/fresco/issues/1472 (getSize and getCount return -1 when app is restarted).

So this PR is not needed.

Sorry,

Eric

PR #1488, opened 2016-09-30
Fix creation of synthetic methods

saantiaguilera:
Fresco is creating a lot of synthetic methods everywhere. Im changing to package access all the ones I found across the dependencies I use daily.

This dependencies are:
- Drawee
- Drawee-pipeline
- Drawee-backends & Imagepipeline-backends (they were empty lol)
- ImagePipeline
- ImagePipeline-OkHttp3
- ImagePipeline-OkHttp

It would be really great to keep this into mind for future code, since in android we have the limitation of the 64k methods dex, and believe it or not, its really easy to reach them :)

You can see them easier with the j2re lint if using android studio =)

### DIFF
--- a/drawee-backends/drawee-pipeline/src/main/java/com/facebook/drawee/backends/pipeline/DraweeConfig.java
+++ b/drawee-backends/drawee-pipeline/src/main/java/com/facebook/drawee/backends/pipeline/DraweeConfig.java
@@ -23,7 +23,7 @@ public class DraweeConfig {
   @Nullable
   private final ImmutableList<DrawableFactory> mCustomDrawableFactories;
 
-  private DraweeConfig(Builder builder) {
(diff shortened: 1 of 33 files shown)


saantiaguilera:
Signed the CLA! :)

kirwan:
I absolutely know how easy it is to reach 64k methods but we've had a discussion about this and I'm afraid we don't think this is a change we want to make.

I appreciate what you've said about synthetic methods but we currently prefer to accept those in favour of clear and correct visibility of class fields and methods.

If you need to enable multidex, it's [simple to do](https://developer.android.com/studio/build/multidex.html). Alternatively, if you're not already using Proguard that will help remove and inline some methods though I admit I've not personally run tests to compare the method counts with different Proguard settings.

kirwan:
I'm going to close this PR but you can still reply to this thread if you have something else to suggest or ask.

PR #1490, opened 2016-10-04
Fixes FIT_CENTER ScaleType example

Widcket:
In the custom ScaleType example the variables scaleX and scaleY where defined as sX and sY, but used as scaleY and scaleX.  Also the class was declared as `static abstract`, which is not possible. This pull request fixes the example.


kirwan:
You're absolutely right. Thanks for fixing this example.

Issue #1492, opened 2016-10-06
Force rotation for image with EXIF orientation tag


nesterov-n:
Fresco 0.14 added new class `RotationOptions.forceRotation(int angle)` allows users to rotate image to custom angle. `forceRotation` doesn't pay attention to EXIF orientation tag. I think it will be more useful to calculate angle passed to `JpegTranscoder` as difference between EXIF orientation  tag and angle set by user.
In my opinion, most common use-case for for force rotation is "Show image rotated by 90 degrees from it's natural orientation", not "Show image rotated by 90 degrees from how it saved in file "
It can be done by custom `PostProcessor` now. But it is additional time-consuming step and additional memory allocation.
If you approve this proposal I can prepare PR for it.

kirwan:
@nesterov-n Your proposal makes perfect sense and a PR would be gratefully received.

erikandre:
Closing issue as the pull request was merged some time ago

PR #1498, opened 2016-10-10
Fix force rotation

nesterov-n:
This PR referrs #1492 
Rotation angle in `ResizeAndRotateProducer` now is computing according to image metadata. It's for rotating image from it's natural orientation.
Add `RotationOprtions.disableRotation()` to disable rotation. Previously it could be done via `RotationOptions.forceRotation(NO_ROTATION)`

### DIFF
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/ResizeAndRotateProducer.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/ResizeAndRotateProducer.java
@@ -292,11 +292,7 @@ public class ResizeAndRotateProducer implements Producer<EncodedImage> {
     if (rotationOptions.useImageMetadata()) {
       return rotationFromMetadata;
     }
-    int angle = rotationFromMetadata + rotationOptions.getForcedAngle();
-    while (angle >= FULL_ROUND) {
-      angle -= FULL_ROUND;
-    }
-    return angle;
+    return (rotationFromMetadata + rotationOptions.getForcedAngle()) % FULL_ROUND;
   }
 
   private static int extractOrientationFromMetadata(EncodedImage encodedImage) {
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/ResizeAndRotateProducer.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/ResizeAndRotateProducer.java
@@ -46,6 +46,7 @@ public class ResizeAndRotateProducer implements Producer<EncodedImage> {
   private static final String ORIGINAL_SIZE_KEY = "Original size";
   private static final String REQUESTED_SIZE_KEY = "Requested size";
   private static final String FRACTION_KEY = "Fraction";
+  private static final int FULL_ROUND = 360;
 
   @VisibleForTesting static final int DEFAULT_JPEG_QUALITY = 85;
   @VisibleForTesting static final int MAX_JPEG_SCALE_NUMERATOR = JpegTranscoder.SCALE_DENOMINATOR;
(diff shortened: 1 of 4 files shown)


nesterov-n:
Seems that build fail isn't caused by my recent changes.
> * What went wrong:
> Execution failed for task ':static-webp:downloadLibpng'.
> java.net.ConnectException: Connection timed out

How can I retry build on CI server?

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/ResizeAndRotateProducer.java
```diff
@@ -284,18 +285,29 @@ private static int getScaleNumerator(
...
-        case RotationOptions.ROTATE_180:
-        case RotationOptions.ROTATE_270:
-          return rotationAngle;
-        default:
-          return 0;
-      }
+      return rotationFromMetadata;
+    }
+    int angle = rotationFromMetadata + rotationOptions.getForcedAngle();
+    while (angle >= FULL_ROUND) {
```
kirwan: Rather than this while loop, you can simply return `angle % FULL_ROUND`

imagepipeline/src/test/java/com/facebook/imagepipeline/producers/ResizeAndRotateProducerTest.java
```diff
@@ -167,8 +167,31 @@ public void testDoesNotTransformIfImageRotationAngleUnkown() {
...
+    provideIntermediateResult(DefaultImageFormats.JPEG);
+    verifyIntermediateResultPassedThroughUnchanged();
+
+    provideFinalResult(DefaultImageFormats.JPEG);
+    verifyFinalResultPassedThroughUnchanged();
+    verifyZeroJpegTranscoderInteractions();
+  }
+
+  @Test
+  public void testDoesNotTransformIfMetadataAngleAndRequestedRotationHaveOppositeValues() {
```
kirwan: 💯

kirwan:
I re-ran the build on Circle CI and it's passed this time.

nesterov-n:
@kirwan Many thanks for your comments. I fixed it. Please review.

kirwan:
I haven't tested but I wonder whether drawee and the pipeline will honour this new option if there is no post-processor.

In [`DecodeProducer.doDecode()`](https://github.com/facebook/fresco/blob/821ebca43184d4c167f87f21a78f599ddc4baef6/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java#L217) we call through to the decoder which for JPEGs passes the rotation from the image into a new ClosableStaticBitmap [with the rotation angle from the encoded image](https://github.com/facebook/fresco/blob/1469512e84394b352405a1d26b92fe4eb5760a8b/imagepipeline/src/main/java/com/facebook/imagepipeline/decoder/DefaultImageDecoder.java#L188).

We can't stop the `EncodedImage` holding the correct value because these are cached for multiple requests so instead I think we might need to add a new property to `ImageDecodeOptions` or otherwise stop the value being passed through.

That doesn't necessarily block this PR but I'm interested to know what you think and whether you could test and if necessary fix that scenario.

nesterov-n:
@kirwan I'm not familiar with pipeline implementation but I can try to reproduce your scenario and check results. Should we hold this PR open until it's checked?

kirwan:
@nesterov-n I'm going to merge this but we would like to ensure the situation I've described works.

Let us know if it's something you'd like to look at otherwise we may add a comment to the Javadocs that this may not work in the situation I've outlined and we'll encourage someone to check/fix it.

PR #1512, opened 2016-10-26
Add largeHeap flag support.

wstczlt:
Summary:
Many applications using largeHeap flag, fresco should obtain more memory for cache in this case.


wstczlt:
Hi, this pull request points to resolve problem of largeHeap flag. Many applications set largeHeap in it's AndroidManifest.xml, application which have this flag will obtain more memory at runtime. For that reason, I think fresco should obtain more memory for cache, because we just want to use 'maxMemory / 4'.

wstczlt:
ping...

wstczlt:
thanks!

kirwan:
Sorry for the slow response from us. I just wanted to explain why we prefer to keep the value we have as the default.

The `MemoryCacheParams` are configurable so this is something you're already free to alter for your purposes, depending on the specific needs of your app.

However we suspect that to take up to a quarter of the large heap size by default would be excessive for the majority of cases.

Feel free to comment further or make other suggestions but for now I'm going to close this pull request.

PR #1541, opened 2016-11-16
update

xuwakao:
update


kirwan:
It looks like you're just trying to merge back changes which are already here so I don't think there's anything for us here.

If you have something specific to offer, please create a specific PR with detailed title and summary.

PR #1544, opened 2016-11-16
Update 02-using-image-pipeline.md

aphexcx:
As mentioned in this [SO post](http://stackoverflow.com/questions/32830745/what-should-callercontext-param-be-for-frescos-prefetchtobitmapcache), the caller Context is allowed to be null.


lambdapioneer:
Thanks for adding to the the documentation :)

PR #1585, opened 2016-12-18
Fix windows unfriendly test

ChrisA89:
Update file path to work correctly when run on windows as well as mac/linux.


ChrisA89:
(I've now signed the cla)

lambdapioneer:
Thanks @ChrisA89, for fixing the test. I'll import the change :)

lambdapioneer:
Hey @ChrisA89, I think that the change is no longer necessary due to 2d3d4b6f098646ff6dd3c748f8af8d7949ecb4ae and c6dcedcfbdd94ee26d2911bdfcef30f812551010.

Can you confirm this on Windows as well?

oprisnik:
I have also fixed another test in 554779490f511db12e122953d9470991805dcfd1.

Thanks again for your contribution, but I'll close this PR since it should not be needed any more.

Issue #1413, opened 2016-08-12
Unknown image format exception


dmitrychistyakov:
Hello, this issue is mentioned earlier, but i created new bug to show pure example of how to reproduce it and why it happens.
I noticed this issue when test my app on device with WiFi disabled and 4G enabled, but there is very important detail. I have negative balance on my phone account, so there is no actually internet on 4G.

So how issue looks like: 
1. I turn off Wifi there is only 4G working
2. I try to load image and got error
3. I enable wifi and try again - image still not loading

That happens, because my 4G internet provider in case of negative balance returns me html page like "please add money to your account" on every html request to show me it when i open some page in browser

The bad thing is that Fresco save this page to cache (I found it there) and tries to decode it as image again and again so user will never see actual image event if internet available.

So, maybe in case of decode exceptions better not to save corrupted data to cache

PR #1586, opened 2016-12-19
#1413: Remove incorrect cached data from caches.

ChrisCanCompute:
I'm not convinced by this change, and would welcome any feedback.

I think that this solution should fix the problem, but would require the library to load the image twice before both caches are cleared. The caches don't talk to each other at the moment, so I'm not sure how to solve that.

This change would also require tests (but I want to get feedback on what I've changed first in case I'm barking up the wrong tree).

### DIFF
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/cache/BufferedDiskCache.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/cache/BufferedDiskCache.java
@@ -187,6 +187,11 @@ public class BufferedDiskCache {
                   try {
                     result = new EncodedImage(ref);
                     result.setEncodedCacheKey(key);
+                  } catch (Exception e) {
+                    // Remove the cached data if it cannot be used to create an encoded image.
+                    CloseableReference.closeSafely(ref);
+                    remove(key);
+                    throw e;
                   } finally {
                     CloseableReference.closeSafely(ref);
                   }
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/EncodedMemoryCacheProducer.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/EncodedMemoryCacheProducer.java
@@ -18,6 +18,8 @@ import com.facebook.imagepipeline.memory.PooledByteBuffer;
 import com.facebook.imagepipeline.request.ImageRequest;
 import com.facebook.cache.common.CacheKey;
 
+import com.android.internal.util.Predicate;
+
 /**
  * Memory cache producer for the encoded memory cache.
  */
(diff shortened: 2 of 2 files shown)


lambdapioneer:
Hi @ChrisA89, thanks for the pull-request.

I'm not sure which underlying problem you are tackling with this change. Can you quickly explain it and describe how one can reproduce it? Alternatively, you can also open a new issue for this. This will certainly help me to provide good feedback to your pull-request :)

ChrisA89:
The aim was to fix #1413 (https://github.com/facebook/fresco/issues/1413).

imagepipeline/src/main/java/com/facebook/imagepipeline/cache/BufferedDiskCache.java
```diff
@@ -187,6 +187,11 @@ public EncodedImage call()
                   try {
                     result = new EncodedImage(ref);
```
lambdapioneer: This line would not throw an `Exception` for corrupted data. The `EncodedImage` is just a wrapper around the encoded image bytes/inputstream.

It is the `DecodeProducer` who would fail during the pipeline processing. The path triggered by bad images is `DecodeProducer#handleError`. I think one possibility would to propagate the failure to the `BaseProducerContext`. Then you are on a higher level where you can find a way to issue a remove to the caches.

ChrisA89: Awesome. I'll look into it. Thanks for the help!

ChrisA89:
`DecodeProducer#handleError` is calling `Consumer#onFailure` already.

I propose that we update the `DiskCacheWriterConsumer` and `EncodedMemoryCacheConsumer` to override `DelegatingConsumer#onFailureImpl` and record that a failure happened, meaning they can switch on this and not store anything in the cache if `Consumer#onNewResult` is called after the failure.

How does that sound?

lambdapioneer:
Hey @ChrisA89,

I am very sorry that I have not responded in a long time. This seems to have fallen between the cracks during the Christmas holidays :/

The tricky part with propagating the issue from the `DecodeProducer` is that the `EncodedMemoryCacheConsumer ` and the `DiskCacheWriterConsumer ` are upstream. One possibility would to catch the error downstream and then issue a removal from the caches.

Having thought about this a bit more, would be a basic sanitizer with the registered `FormatChecker`s just when we receive the first bytes from the `NetworkFetcher`. Then we can abort the request right away. Also we don't remove things from the `DiskCache` when we try to store the "bad data". (@oprisnik knows more about this)

I think one good implementation for this would to add a `InputSanitizerProducer` right after the `NetworkFetchProducer`/`LocalFileFetchProducer`. This obviously would be a bit more code to write (you can find an example of adding a new producer here: b9f6fdbc9b02971c7ef99bd4ceec7a48f8c196c7 and 4bc362217390e311a25f1f0d1b390c0bbdd9f9b7)

Let me know if you are still interested on taking on this :)

erikandre:
Thanks for the contribution @ChrisCanCompute 
I'm going to close this now as from what I can see, the pull request is not in a state where we can merge it and there has been no activity for quite some time.

PR #1607, opened 2016-12-29
Fix WebpBitmapFactoryImpl.createBitmap() being called with wrong class

korDen:
When JNI debugging is enabled, ART is very pedantic about what class you are passing when invoking static method. In this case, this is causing a crash:

runtime.cc:476] JNI DETECTED ERROR IN APPLICATION: can't call static android.graphics.Bitmap com.facebook.webpsupport.WebpBitmapFactoryImpl.createBitmap(int, int, android.graphics.BitmapFactory$Options) with class java.lang.Class<android.graphics.Bitmap>
runtime.cc:476]     in call to CallStaticObjectMethodV
runtime.cc:476]     from android.graphics.Bitmap com.facebook.webpsupport.WebpBitmapFactoryImpl.nativeDecodeByteArray(byte[], int, int, android.graphics.BitmapFactory$Options, float, byte[])


lambdapioneer:
Nice catch 👍  Will import and merge after testing

PR #1630, opened 2017-01-17
Fix media variations not working if candidate not a jpeg image

FtheGFW:


kirwan:
Thanks for this contribution @FtheGFW.

kirwan:
@FtheGFW The changes you've made to `MediaVariationsIndexDatabase` are clearly worthwhile but I don't think related to the non-jpeg case. Would you mind moving those into a new PR (rebased on master to resolve the conflict)?

As for the issue with non-jpeg images, such images shouldn't go through the progressive jpeg flow, so I don't think this is where we'd want to administer a fix.

I think I know why this happens (a non-final image was previously always assumed to be a jpeg) but we should fix that assumption rather than making any changes to jpeg decoding. We should change the format of a couple of the variations in the sample app to test these cases.

I'm working on media variations changes this week so if you aren't able to make the changes I'll get to it. I'm very happy for you to fix the issue if you want to do it though.

FtheGFW:
@kirwan I agree with your points about the non-jpeg image issue, so I committed a new [PR](https://github.com/facebook/fresco/pull/1672) to fix duplicate database entry bug

Issue #1633, opened 2017-01-19
SimpleDraweeSpanTextView does not resize when width is match_parent


angarron:
When `SimpleDraweeSpanTextView`'s width is set to `match_parent` and a `DraweeSpan` is displayed which is larger than the view's height, going back to text does not resize the view.

See gif (pardon the horrendous compression):

![output](https://cloud.githubusercontent.com/assets/8857134/22095144/4eb50b72-ddc7-11e6-9a64-6c93be074a2a.gif)

Here is an example of offending code (just stuffed inside of an activity, nothing complicated, and don't forget to set the width to `match_parent` on the `SimpleDraweeSpanTextView`):

<details>

```
        textView = (SimpleDraweeSpanTextView) findViewById(R.id.textview);
        textButton = (Button) findViewById(R.id.text_button);
        imageButton = (Button) findViewById(R.id.image_button);

        textButton.setOnClickListener(new View.OnClickListener() {
            @Override
            public void onClick(View v) {
                final DraweeSpanStringBuilder builder = new DraweeSpanStringBuilder();
                builder.append("hello");
                textView.setDraweeSpanStringBuilder(builder);
            }
        });

        imageButton.setOnClickListener(new View.OnClickListener() {
            @Override
            public void onClick(View v) {
                final DraweeSpanStringBuilder builder = new DraweeSpanStringBuilder();
                final int startIndex = builder.length();
                builder.append("#");

                final DraweeController controller = Fresco.newDraweeControllerBuilder()
                    .setUri("https://cdn.discordapp.com/emojis/232595451734392832.png")
                    .build();

                final DraweeHierarchy draweeHierarchy = GenericDraweeHierarchyBuilder.newInstance(getResources())
                    .setPlaceholderImage(new ColorDrawable(Color.TRANSPARENT))
                    .build();

                builder.setImageSpan(
                    MainActivity.this,
                    draweeHierarchy,
                    controller,
                    startIndex,
                    builder.length() - 1,
                    dpToPx(30),
                    dpToPx(30),
                    true,
                    DraweeSpan.ALIGN_BASELINE);

                textView.setDraweeSpanStringBuilder(builder);
            }
        });

```

</details>

[...]

PR #1634, opened 2017-01-19
Use BufferType.SPANNABLE in setText of SimpleDraweeSpanStringBuilder

angarron:
Explicitly uses `BufferType.SPANNABLE` to let `SimpleDraweeSpanTextView` know that it might need to resize itself. See associated issue: https://github.com/facebook/fresco/issues/1633

Feel free to suggest alternative solutions; I think this is safe to do and a cursory search didn't turn up anything better, but I suspect maintainers of `DraweeSpan` might know more about spans than I do!


oprisnik:
Thank you very much @angarron, looks good to me!

Issue #1513, opened 2016-10-26
Fresco and okHttp


rooton:
Hi!

I try to use OkHTTP3 with Fresco.

**Application:**

```
public class MyApplication extends Application {

       public MyApplication () {
          super();
       }

    @Override
    public void onCreate() {
        super.onCreate();

        OkHttpClient okHttpClient = new OkHttpClient();

        ImagePipelineConfig config = OkHttpImagePipelineConfigFactory
                .newBuilder(this, okHttpClient)
        .build();

        Fresco.initialize(this, config);
    }
}
```

**Gradle:**

```
compile 'com.facebook.fresco:fresco:0.14.1'
compile 'com.squareup.okhttp3:okhttp:3.4.1'
compile 'com.facebook.fresco:imagepipeline-okhttp3:0.14.1'
```

**BaseAdapter (ListView) / getView**

```
holder.simpleDrawerView= (SimpleDraweeView) inList.findViewById(R.id.simpleDrawerView);
. . .

Uri uri = Uri.parse(getItem(pos)); //img url
//for example: Uri uri = Uri.parse("http://thumbs.fullhd.pics/350x0/974560d3c39efda1ffd3163d9df4019b/"+System.currentTimeMillis());
holder.simpleDrawerView.setImageURI(uri);
```

**Active scrolling, sometimes gives an error:**

> FATAL EXCEPTION: OkHttp Dispatcher
>                                                                         java.lang.IllegalArgumentException: in == null
>                                                                             at okio.Okio.source(Okio.java:127)
>                                                                             at okio.Okio.source(Okio.java:203)
>                                                                             at okhttp3.internal.connection.RealConnection.connectSocket(RealConnection.java:191)
>                                                                             at okhttp3.internal.connection.RealConnection.buildConnection(RealConnection.java:173)
>                                                                             at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:114)
>                                                                             at okhttp3.internal.connection.StreamAllocation.findConnection(StreamAllocation.java:193)
>                                                                             at okhttp3.internal.connection.StreamAllocation.findHealthyConnection(StreamAllocation.java:129)
>                                                                             at okhttp3.internal.connection.StreamAllocation.newStream(StreamAllocation.java:98)
>                                                                             at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:42)
[...]

PR #1647, opened 2017-02-01
Improve error handling for okhttp

lucas34:
PR for issue #1513 

I added the fetchWithRequest() method inside Okhttp2 fetcher to make the implementation more consistent.


lambdapioneer:
Great work! Thanks for fixing this - will import and get it reviewed 👍

PR #1672, opened 2017-02-22
fix duplicate media variant entries

FtheGFW:
**motivation** :
when I download some images with media variations feature enabled, then I got some entries written to the media variations index database, but if I cleared all the previous cached  images (or load the same image while writing index info to the db) , and reload those images again, I would got duplicate entries with the same mediaId, width, height and resourceId pointing to the same disk cache key
   
**reproduce steps** :
0. check out current master branch
1.  just run example/media-variations
2. click to load all the images
3. then click the "Clear cache and reset" menu
4. reload all those images again
5. repeat step 3 and 4 if you want
6. pull the media-variations index database from the test device, open it using a sqlite3 client and query all entries from media_variations_index table
7. ***observable result*** : you would find more than one entries with identical mediaId, width, height, resourceId, cache_key and resource_id (except the auto generated primary key)

**Test plan**
execute the reproduce steps based on this pr,
check out if there any duplicates entries, if no two entries are identical then the issue is fixes.

### DIFF
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/MediaVariationsIndexDatabase.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/MediaVariationsIndexDatabase.java
@@ -35,6 +35,7 @@ import com.facebook.imagepipeline.request.MediaVariations;
 import bolts.Task;
 
 public class MediaVariationsIndexDatabase implements MediaVariationsIndex {
+
   private static final String TAG = MediaVariationsIndexDatabase.class.getSimpleName();
 
   private static final String[] PROJECTION = {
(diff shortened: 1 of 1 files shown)


kirwan:
The changes look fine but can you please update the summary in line with the template? Then I can import it.

FtheGFW:
@kirwan  BTW, would adding a method named clearCachedVariants() additionally to MediaVariationsIndex interface and invoking it  immediately after disk caches were cleared be a litter bit better ?

kirwan:
Yes @FtheGFW, that's coming. This is why the index is still in development and behind an experimental config flag but I'm very happy to have people like you looking at it and making suggestions and improvements.

kirwan:
I've updated the initial comment as that's included in the commit message when this is merged.

Issue #1676, opened 2017-02-24
setRetainImageOnFailure does not work IncreasingQualityDataSourceSupplier when main image process failed immediately


yunarta:
This if how to make the test

1. Create a builder with low res and high res images in the request
2. Enable retain image on failure
3. The high res image is actually pointing to URL which will gives error instead of the image
4. Then when the high res image is failing, the low res image is not being displayed as well

This is caused by the result from the first image being reported here immediately [here](https://github.com/facebook/fresco/blob/84aeab6895f3177cee473e85656cf898112f263f/fbcore/src/main/java/com/facebook/datasource/IncreasingQualityDataSourceSupplier.java#L171)

By delaying the error reporting until all the image download process is finished, then we will have the last image downloaded

I'm making fork of this and will send pull request soon

yunarta:
Now i'm also seeing new problem coming from consumer and producer racing conditions.

In IncreasingQualityDataSourceSupplier the data source are being subscribed in CallerThreadExecutor as shown [here](https://github.com/facebook/fresco/blob/84aeab6895f3177cee473e85656cf898112f263f/fbcore/src/main/java/com/facebook/datasource/IncreasingQualityDataSourceSupplier.java#L102)

As per my debugging, failure may come from OkHttp thread.

In my pull request, i delayed the notification of exception so it will allow AbstractDraweeController to consume any results from the IncreasingQualityDataSourceSupplier.

However now i found that if setFailure is to be called immediately after setResult, AbstractDraweeController will consume one of the increasing quality result, but the dataSource state [here](https://github.com/facebook/fresco/blob/ad4e2ef2550efea7c9304d889c24dd574d1767b8/drawee/src/main/java/com/facebook/drawee/controller/AbstractDraweeController.java#L470) had changed to finished.

This is due to the DataSource stated had been modified [here](https://github.com/facebook/fresco/blob/84aeab6895f3177cee473e85656cf898112f263f/fbcore/src/main/java/com/facebook/datasource/AbstractDataSource.java#L238) before notification had been consumed.

Most of the producer consumer problem can be solve easily with RxJava, i'll be looking forward of next Fresco version created based on RxJava :)

PR #1677, opened 2017-02-24
Fix for #1676, setRetainImageOnFailure does not work IncreasingQualit…

yunarta:
By delaying the error until all the image data source is completed, now retain image on failure is working, however now the side effect is the retry image is not drawn.

Should the retry image to be drawn above failed image?


yunarta:
Seems that the test failed due to new changes that will send all the increasing quality result regardless of the highest quality failed state.

However this is required is set retain images to be fixed.

This situation actually can be solve easily with Reactive programming, since the caller can define whether they wanted to delay the error or not.

Any advice?

foghina:
Hi @yunarta, I landed a similar version of your PR in 62c58fd, that includes a test fix. Closing this now.

Issue #1685, opened 2017-03-02
Dynamically link jpegturbo lib for reduce library size.


desmond1121:
I've noticed native lib **jpegturbo** is both statically linked in bellowing module:

- **imagepipeline** (See  [imagepipeline/Android.mk#L28](https://github.com/facebook/fresco/blob/master/imagepipeline/src/main/jni/imagepipeline/Android.mk#L28)) 
- **static-webp** (See [static-webp/Android.mk#L37](https://github.com/facebook/fresco/blob/master/static-webp/src/main/jni/static-webp/Android.mk#L37))

It's size is 165KB kb in armeabi-v7 and 409KB in x86 (so big!), would cause a lot of redundant apk size!

Since every body who use **static-webp** library would and must already have  **imagepipeline** in their app, so I think we can do some trick:

- make fb_jpegturbo as a shared library and dynamically link them in two modules;
- after ndk-build success in **static-webp**, delete fb_jpegturbo.so in this module otherwise it would conflict with **imagepipeline**'s.

If this idea is accepted, I would give a PR in several days.

lambdapioneer:
Hi @desmond1121, thanks for raising this issue! Having both projects statically linking against a larger library sounds bad and we should fix this. Your solution sounds good and hopefully there's an easy way to do this.

Definitely looking forward to read your PR! 👍 Would great to have some numbers on the size win.

desmond1121:
@lambdapioneer PR #1686

PR #1686, opened 2017-03-02
Build jpegturbo as a shared library.

desmond1121:
Link jpegturbo as a shared library for reduce apk size.

See [#1865](https://github.com/facebook/fresco/issues/1685)

After this work, if somebody use webp support, his so library would be (file size is measured in armeabi-v7a abi):

- **libfb_jpegturbo.so** 165kb
- **libimagepipeline.so** 26kb
- **libstatic-webp.so** 290kb

Totally: 481kb

Before (which would link jpegturbo as a static library):

- **libimagepipeline.so** 230kb
- **libstatic-webp.so** 362kb

Totally: 592kb (So we reduce 117kb just in armeabi-v7 !)

But if someone only use imagepipeline (not use webp support), he would also have 39kb size reduce!


lambdapioneer:
Cool, looks good - I'll import it. Can you perhaps add here how you tested that the webp support still works correctly?

desmond1121:
Sorry @lambdapioneer , there is some mistake in my previews measurement: after optimize, libimagepipeline.so's size in armeabi-v7 is 94kb (not 26kb), and size reducement in armeabi-v7 would be 43kb.

Test plan:
Since `static-webp` module exclude libjpeg_turbo.so, so I extract webp test to a new module : `static-webp-test`, it would rely on `imagepipeline` and `static-webp`.  (see [43c9f91](https://github.com/facebook/fresco/commit/43c9f91ab0f96ea5004df8758bb979e9bf6ab6ac))
I've ran test `WebpBitmapFactoryTest` and `WebpDecodingTest` and all test case is passed.

lambdapioneer:
Hi @desmond1121, since this change increased the APK size if just using core library, I'm going to revert it. We are committed to have a slim Fresco core that has minimal size impact on the developers' apps. Also only 10% of Fresco users download the webp-support package. CC @kirwan

PR #1691, opened 2017-03-07
Extract static-webp-test.

desmond1121:
See #1686 , since static-webp module exclude libjpeg_turbo.so, so I extract webp test to a new module : static-webp-test, it would rely on imagepipeline and static-webp.


desmond1121:
@lambdapioneer I've update ci script and it passed then:)

kirwan:
As the other pull request was reverted, I'm going to close this too. Thanks for the contribution though @desmond1121.

PR #1692, opened 2017-03-07
Fix two links

paour:
Extra spaces are likely causing the links to be rendered incorrectly on the main docs site (GH is more lenient).

datasources_and_datasubscribers_-_fresco


lambdapioneer:
@paour, thanks for fixing this!

Issue #1518, opened 2016-10-31
Android Studio 2.2.2 After the introduction of the XML version of the 0.10.1 property without prompt


bjxunpai:
Android Studio 2.2.2 After the introduction of the XML version of the 0.10.1 property without prompt
http://share.weiyun.com/678deb081fceabb0f111400eca9c0d13

kirwan:
Sorry, I'm not sure what you're asking. I wouldn't expect Android Studio to auto-suggest the Fresco custom properties.

Just to note: you've called the namespace "fresco" but this will provide all properties defined in your app or other libraries you use.

bjxunpai:
when I use the Fresco 0.10.1，there  appears the custom property. But when I use the Fresco 0.14.1，no custom property prompts. How can I solve this problem?

shenjingfs:
Same use Android Studio 2.2.2, I want Android Studio to auto suggest Fresco custom properties, not just the following：
> * Fresco 0.10.0
![image](https://cloud.githubusercontent.com/assets/12003831/19923286/9d33ad90-a121-11e6-821b-0feb2a789a01.png)
![image](https://cloud.githubusercontent.com/assets/12003831/19923301/a5201e30-a121-11e6-9d66-e2e192a10a26.png)
> * If I use Fresco 0.14.1，Android Studio auto suggest a new property：
![image](https://cloud.githubusercontent.com/assets/12003831/19923359/eedeabe0-a121-11e6-9492-989cf2cc83f5.png)
![image](https://cloud.githubusercontent.com/assets/12003831/19923314/b01d735a-a121-11e6-85fd-e492e96ac939.png)

Or should I ask questions  on Stack Overflow？

kirwan:
The actualImageUri attribute was added more recently than 0.10.0 so it's no surprise that doesn't appear if that's the version you're using.

I've just found this Stack Overflow question about [losing custom attribute autocomplete](http://stackoverflow.com/questions/30684613/android-studio-xml-editor-autocomplete-not-working-with-support-libraries). If that's what you're experiencing, there are a couple of suggested answers.

bjxunpai:
> `https://cloud.githubusercontent.com/assets/12003831/19923314/b01d735a-a121-11e6-85fd-e492e96ac939.png`

shenjingfs:
Thank you for your help. I have seen this problem, and in my Android Studio auto complete looks working:
[...]

PR #1717, opened 2017-04-08
#1518 add parent for SimpleDraweeView in attrs.xml to make auto-compl…

nesterov-n:
Made `GenericDraweeHierarchy` as parent of `SimpleDraweeView` in `attrs.xml`
I believe, what it fixes auto-complete problem from #1518

### DIFF
--- a/drawee/src/main/res/values/attrs.xml
+++ b/drawee/src/main/res/values/attrs.xml
@@ -124,7 +124,7 @@
 
   </declare-styleable>
 
-  <declare-styleable name="SimpleDraweeView">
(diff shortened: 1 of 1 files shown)


lambdapioneer:
Hi @nesterov-n, I've tried to reproduce your fix, but it does not work in Android Studio (v2.3) for me. With which one have you tested it?

nesterov-n:
@lambdapioneer To be honest, I did not test this particular fix. I fixed same problem in my own project this way recently. I hastily suggested that trick with `parent` in `attrs.xml` would work in this case too.
Now I see that it doesn't work. Will try to figure out problem and open new PR.
Sorry for inconvenience.

PR #1720, opened 2017-04-12
[samples] Use more varied images in the RecyclerView sample

mkonicek:

## Motivation (required)

Currently the example has only a few images which makes it hard to visually debug what's in cache:

screenshot 2017-04-12 14 09 40

This PR makes the example a bit more realistic by using many different images (120 in total):

screenshot 2017-04-12 13 53 44

## Test Plan (required)

Ran the Showcase app from Android Studio, see the screenshot above.


lambdapioneer:
Cool, didn't know lorempixel.com before. I think it's fine since we are able to revert it back in case that service shuts down for any reason.

PR #1722, opened 2017-04-16
change setBitmapConfig return type from void to ImageDecodeOptionsBui…

voiddog:

## Motivation (required)

Other set method return the builder but setBitmapConfig not return.

## Test Plan (required)

I think this does not need to be tested


lambdapioneer:
Thanks for fixing this! Will land this now :)

PR #1728, opened 2017-04-19
Optimize BitmapDrawable, add comments

mkonicek:

## Motivation (required)

Here are the takeaways from a recent experiment at fb related to drawing circular bitmaps:

It's probably possible to optimize the rounding drawables used by Fresco: 
1. Let's not apply a matrix if we don't need to
2. Consider avoiding a path when using addRoundRect alone would work

This PR does those 2 optimizations. For 2., we skip paths altogether and call `Canvas.drawCircle`, but only on Android versions where it's faster. From the research:

"On Android 11-16, drawRoundRect is actually slow because those api versions used alpha mask textures to draw shapes.. Consequently, the support library falls back to doing 7 draw commands on those api versions (drawing arcs and rects)."

(Also add some comments - it took me a while to understand the purpose of each class, hopefully this will save some time for people touching the code in the future.)

## Test Plan (required)

Ran the Showcase app, the Rounded Corners example looks the same as before, also when changing scale types:

screenshot 2017-04-19 18 11 10
screenshot 2017-04-19 18 11 17

screenshot 2017-04-19 18 10 35
screenshot 2017-04-19 18 10 44


drawee/src/main/java/com/facebook/drawee/drawable/RoundedBitmapDrawable.java
```diff
@@ -270,7 +298,10 @@ private void updateTransform() {
       mIsShaderTransformDirty = true;
       mParentTransform.invert(mInverseParentTransform);
       mTransform.set(mParentTransform);
-      mTransform.preConcat(mBoundsTransform);
+      if (!mBoundsTransform.isIdentity()) {
```
oprisnik: Are you sure that this is worth doing? `preConcat` should end up in https://github.com/google/skia/blob/master/src/core/SkMatrix.cpp#L715 which already does the same check internally.

drawee/src/main/java/com/facebook/drawee/drawable/RoundedBitmapDrawable.java
```diff
@@ -322,7 +345,10 @@ private void updatePaint() {
       mIsShaderTransformDirty = true;
     }
     if (mIsShaderTransformDirty) {
-      mPaint.getShader().setLocalMatrix(mTransform);
+      // Draw the bitmap at the correct scale
+      if (!mTransform.isIdentity()) {
+        mPaint.getShader().setLocalMatrix(mTransform);
```
oprisnik: Do you have to reset the matrix otherwise? I.e. does this work if you switch from a complex to the circular case?

samples/showcase/src/main/java/com/facebook/fresco/samples/showcase/drawee/DraweeRoundedCornersFragment.java
```diff
@@ -170,11 +170,12 @@ private void changeDraweeViewScaleType(
     hierarchy.setActualImageFocusPoint(focusPoint != null ? focusPoint : new PointF(0.5f, 0.5f));
 
     final RoundingParams roundingParams = Preconditions.checkNotNull(hierarchy.getRoundingParams());
-    if (BITMAP_ONLY_SCALETYPES.contains(scaleType)) {
-      roundingParams.setRoundingMethod(RoundingParams.RoundingMethod.BITMAP_ONLY);
-    } else {
-      roundingParams.setOverlayColor(mWindowBackgroundColor);
-    }
+//    if (BITMAP_ONLY_SCALETYPES.contains(scaleType)) {
```
oprisnik: Can you remove the comment please :)

mkonicek: Oops! Fixed.

PR #1729, opened 2017-04-19
#1518 declare custom attributes against SimpleDraweeView 

nesterov-n:
Related to #1518

Declare custom attributes against `SimpleDraweeView`to make auto-complete works.

Checked in AndroidStudio 2.3.1
screen shot 2017-04-19 at 21 53 07

### DIFF
--- a/drawee/src/main/res/values/attrs.xml
+++ b/drawee/src/main/res/values/attrs.xml
@@ -124,12 +124,87 @@
 
   </declare-styleable>
 
-  <declare-styleable name="SimpleDraweeView">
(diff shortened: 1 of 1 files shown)


kirwan:
Sorry for the delay in response.

Is there no way to avoid the duplication? If not, I think it would be better to skip all the comments second time around and just include them in a single, large block. They're already documented within `GenericDraweeHierarchy`.

kirwan:
I'm importing this so it can be included in our imminent release. I'll remove the extraneous comments though.

PR #1734, opened 2017-04-24
[DO NOT MERGE] Demo rounding using View.setClipToOutline(), available on API 21+

mkonicek:
Shows how to use the API 21 method [`View.setOutlineProvider`](https://developer.android.com/reference/android/view/View.html#setOutlineProvider(android.view.ViewOutlineProvider)) to implement rounded corners.

This method supports circles, ellipses and round rectangles with uniform corner radius (documented [here](https://developer.android.com/reference/android/graphics/Outline.html#canClip()) and tried [here](http://stackoverflow.com/questions/43548063/android-view-outline-using-a-custom-path)).

Result (Fresco showcase app):

screenshot 2017-04-24 13 55 25

### DIFF
--- a/drawee/src/main/java/com/facebook/drawee/generic/RoundingParams.java
+++ b/drawee/src/main/java/com/facebook/drawee/generic/RoundingParams.java
@@ -36,10 +36,15 @@ public class RoundingParams {
      * {@link ScalingUtils.ScaleType#CENTER_CROP}, {@link ScalingUtils.ScaleType#FOCUS_CROP} and
      * {@link ScalingUtils.ScaleType#FIT_XY}.
      */
-    BITMAP_ONLY
+    BITMAP_ONLY,
+
+    /**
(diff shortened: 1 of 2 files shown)


drawee/src/main/java/com/facebook/drawee/generic/RoundingParams.java
```diff
@@ -36,10 +36,15 @@
...
+    BITMAP_ONLY,
+
+    /**
+     * On Android versions >= 21 (Lollipop) uses the {@code View.setClipToOutline} method.
+     */
+    OUTLINE,
   }
 
-  private RoundingMethod mRoundingMethod = RoundingMethod.BITMAP_ONLY;
+  private RoundingMethod mRoundingMethod = RoundingMethod.OUTLINE;
```
mkonicek: Just a demo, normally we'd check API level.

mkonicek:
(Just showing how to use the API. Closing this PR.)

PR #1736, opened 2017-04-25
Add a new rounding method using View.setOutlineProvider

mkonicek:
## Motivation

Adds a new rounding method using [`View.setOutlineProvider`](https://developer.android.com/reference/android/view/View.html#setOutlineProvider(android.view.ViewOutlineProvider)) to implement rounded corners. Needs API level 21+.

The motivation is to try out whether this improves scroll performance.

Added a new dropdown to the example in the Showcase app to allow people to try out the outline method.

screenshot 2017-04-25 18 25 51

## Limitations

The main limitation is that the new API `View.setOutlineProvider` must be used with Android Views. To use the outline rounding, users must call `draweeView.setRoundingParams(myRoundingParams)`.

The other rounding methods in Fresco work directly with Drawables, and users call `hierarchy.setRoundingParams(myRoundingParams)`.

### Do we want to merge this?

It won't be possible to use this new rounding method with Litho because Litho doesn't use Android Views. Since Litho is used a lot in the Facebook app, this pull request might actually have little benefit for fb. It is a bit of extra code to maintain. Is it worth adding this to Fresco?

## TODO

Borders are not supported yet. Pausing work on borders until we decide whether we want to merge this.

## Test Plan

Ran the Showcase app. Outline rounding works with all scale types:

screenshot 2017-04-25 18 24 07

The outline API only supports ellipses and rounded rectangles with uniform corner radius (documented [here](https://developer.android.com/reference/android/graphics/Outline.html#canClip()) and tried [here](http://stackoverflow.com/questions/43548063/android-view-outline-using-a-custom-path)). Therefore non-uniform border radii are not supported:

screenshot 2017-04-25 18 26 53

### DIFF
--- a/drawee/src/main/java/com/facebook/drawee/generic/RoundingParams.java
+++ b/drawee/src/main/java/com/facebook/drawee/generic/RoundingParams.java
@@ -36,7 +36,12 @@ public class RoundingParams {
      * {@link ScalingUtils.ScaleType#CENTER_CROP}, {@link ScalingUtils.ScaleType#FOCUS_CROP} and
      * {@link ScalingUtils.ScaleType#FIT_XY}.
      */
-    BITMAP_ONLY
+    BITMAP_ONLY,
+
+    /**
+     * On Android versions >= 21 (Lollipop) uses the {@code View.setClipToOutline} method.
+     */
+    OUTLINE
   }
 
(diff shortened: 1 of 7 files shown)


froxCZ:
+1

kirwan:
I'm closing our old PRs from internal people. We may still pursue this but right now this is just stale.

PR #1737, opened 2017-04-26
Support ScaleType in DraweeSpanStringBuilder

angarron:
Adds support for `ScaleType.FIT_START` to `DraweeSpanStringBuilder`. Maintains original API and specifies previous (now default) behavior as `ScaleType.CENTER`.

## Motivation

A common use case (especially in chat apps when dealing with emojis) is to display images in their native aspect ratio without letting them grow indefinitely.

Below is a screenshot where the first line of text uses `ScaleType.CENTER` and the second line of text uses `ScaleType.FIT_START`.

fresco_screenshot

Other scale types should be easy to add as needed, but I currently only need `ScaleType.FIT_START` so I would like to get feedback on this implementation before supporting other scale types.

## Test Plan

Wrote sample app that tests new behavior (see screenshot), tests still run as expected.

## Next Steps

- [ ] Some concessions were made in order to keep the current (1.3.0) API intact. I believe that `enableResizing` should be dropped everywhere as it introduces ambiguity, and instead I should directly reference `ScaleType.FIT_XY`. However this breaks the current API and I want Fresco maintainers' feedback on that before making a breaking change.
- [ ] I have implemented the computation of new bounds myself, though recently I realized I should instead be trying to funnel that computation through `ScalingUtils` instead. This should be addressed before merging -- as-is, those computations do not have tests so I would like to instead piggyback off of the battle-tested `ScalingUtils`.

Thank you for considering this PR and all feedback is of course appreciated :)

### DIFF
--- a/drawee-span/src/main/java/com/facebook/drawee/span/DraweeSpanStringBuilder.java
+++ b/drawee-span/src/main/java/com/facebook/drawee/span/DraweeSpanStringBuilder.java
@@ -13,13 +13,14 @@ import android.content.Context;
 import android.graphics.Rect;
 import android.graphics.drawable.Animatable;
 import android.graphics.drawable.Drawable;
+import android.support.annotation.Nullable;
+import android.support.v4.util.Pair;
 import android.text.SpannableStringBuilder;
 import android.view.View;
 
 import com.facebook.common.internal.Preconditions;
 import com.facebook.common.internal.VisibleForTesting;
 import com.facebook.common.lifecycle.AttachDetachListener;
-
 import com.facebook.drawee.controller.AbstractDraweeController;
 import com.facebook.drawee.controller.BaseControllerListener;
 import com.facebook.drawee.interfaces.DraweeController;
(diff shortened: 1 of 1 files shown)


drawee-span/src/main/java/com/facebook/drawee/span/DraweeSpanStringBuilder.java
```diff
@@ -299,43 +326,96 @@ public DrawableChangedListener(DraweeSpan draweeSpan, boolean enableResizing) {
...
+          topLevelDrawable.setBounds(0, 0, scaledDimens.width, scaledDimens.height);
 
           if (mDraweeSpanChangedListener != null) {
             mDraweeSpanChangedListener.onDraweeSpanChanged(DraweeSpanStringBuilder.this);
           }
         }
       }
     }
+
+    private Dimens getScaledDimensCenter(final Drawable drawable, final ImageInfo imageInfo) {
```
angarron: I believe I can funnel this computation through `ScalingUtils` instead of computing it here myself -- that work is underway.

oprisnik:
Thank you very much for the PR @angarron. We'll take a look soon :)

angarron:
Good to hear, thanks. I've been a little busy so I haven't made the switch to `ScalingUtils` yet but if I hear that this general approach is good, I'll pump the gas on that a little.

angarron:
Woah awesome to hear. Will verify this works for me and get back to you ASAP.

oprisnik:
Sample code is here: 6ee00684cec93418c8b71e9285d30026c496d3cd

angarron:
@oprisnik looks like this works for us. Thanks for the investigation and the help!

oprisnik:
Great to hear that it worked :)

PR #1761, opened 2017-05-11
fix partial bitmap got displayed

FtheGFW:
## Motivation 
when loading an image for the first time with media-variations enabled, decode producer could got a set of partial images but of none progressive jpeg image format (i.e. webp) continuously from network fetch producer if **under poor network condition**, these partial bitmaps would be treated like progressive jpegs , however a none fully decoded webp image is not useful to the user as the bitmap used to render this partial image may contains lots of pixels not set (observed as pure black or white)
partial_webp

## Test Plan
use web debugging tools like Charles to simulate a pool network conditions, and loading a none jpeg image, make sure no partial bitmap got displayed

### DIFF
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java
@@ -418,8 +418,12 @@ public class DecodeProducer implements Producer<CloseableReference<CloseableImag
     @Override
     protected synchronized boolean updateDecodeJob(EncodedImage encodedImage, boolean isLast) {
       boolean ret = super.updateDecodeJob(encodedImage, isLast);
-      if (!isLast && EncodedImage.isValid(encodedImage) &&
-          encodedImage.getImageFormat() == DefaultImageFormats.JPEG) {
+      if (!isLast && EncodedImage.isValid(encodedImage)) {
+        if (encodedImage.getImageFormat() != DefaultImageFormats.JPEG) {
+          // decoding a partial image data of none jpeg type and displaying that result bitmap
+          // to the user have no sense.
+          return ret && isNotPartialData(encodedImage);
+        }
         if (!mProgressiveJpegParser.parseMoreData(encodedImage)) {
           return false;
         }
@@ -437,6 +441,10 @@ public class DecodeProducer implements Producer<CloseableReference<CloseableImag
         mLastScheduledScanNumber = scanNum;
       }
       return ret;
+    }
+
+    private boolean isNotPartialData(EncodedImage encodedImage) {
+      return encodedImage.getEncodedCacheKey() != null;
     }
 
     @Override


imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java
```diff
@@ -418,8 +418,12 @@ public NetworkImagesProgressiveDecoder(
     @Override
     protected synchronized boolean updateDecodeJob(EncodedImage encodedImage, boolean isLast) {
       boolean ret = super.updateDecodeJob(encodedImage, isLast);
-      if (!isLast && EncodedImage.isValid(encodedImage) &&
-          encodedImage.getImageFormat() == DefaultImageFormats.JPEG) {
+      if (!isLast && EncodedImage.isValid(encodedImage)) {
+        if (encodedImage.getImageFormat() != DefaultImageFormats.JPEG) {
+          // decoding a partial image data of none jpeg type and displaying that result bitmap
+          // to the user have no sense.
+          return ret && isNotPartialData(encodedImage);
```
foghina: Instead of checking `isNotPartialData()`, why not just check `isLast`? So this could be:

```java
return ret && isLast;
```

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java
```diff
@@ -439,6 +443,10 @@ protected synchronized boolean updateDecodeJob(EncodedImage encodedImage, boolea
       return ret;
     }
 
+    private boolean isNotPartialData(EncodedImage encodedImage) {
```
foghina: I don't really like this method name. Instead of having an `isNot` name I'd rather have e.g. `isPartialData()` and combine with `!` at the usage site.

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java
```diff
@@ -439,6 +443,10 @@ protected synchronized boolean updateDecodeJob(EncodedImage encodedImage, boolea
...
+      return encodedImage.getEncodedCacheKey() != null;
```
foghina: I'm not sure this is a reliable way to tell whether this image is partial or not. Can you show why you think this is correct, or use `isLast` as I mentioned above?

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java
FtheGFW: at this point , {isLast} is always false

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java
FtheGFW: Can't use the ```isLast``` here, as this encoded image could came from disk cache produced by media-variations fallback producer, in that case, value of ```isLast``` could be true if the cached image is large enough than the request specified, or false if a full network fetch is needed.
Most importantly, only when ```isLast == false``` will this method be executed, because if it is a last result of a request, the problem I described in the PR should never occur.

FtheGFW: an intermediate result either came from network fetcher under a poor network connection, or disk cache when media-variants could be found with the media-id set on a request, as fresco only cache a image onto disk cache if it received the full raw bytes of the image( or isLast == true, checkout ```DiskCacheWriteConsumer```) and any encoded image came from MediaVariationsFallbackProducer with a media-variant satisfies the request then it should has a valid cache key, you cant checkout DiskCacheReadProducer

foghina:
@kirwan can you take a look here, since you're more familiar with media variations?

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java
kirwan: The existence of the cache key definitely shouldn't be used to determine the completeness of the image data.

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/DecodeProducer.java
kirwan: @FtheGFW is right about this, isLast wouldn't be sufficient.

However, I'm just making a change that will mean we can send back some other flags which may be useful here.

kirwan:
@erikandre just pushed 513fdf32ebb521fc7905b468f72f5f7f36b2af2f which touches some of the same methods. It was attempting to fix a slightly different problem though and I can still reproduce your problem.

We've made a change to the Consumer API so that they now receive an int containing multiple status flags and that may be useful to fix this issue too. A placeholder from the `MediaVariationsFallbackProducer` will now include `IS_PLACEHOLDER`. I wonder if that helps you with this. For non-JPEGs we can initiate decoding only if either `IS_LAST` or `IS_PLACEHOLDER` is et.

@FtheGFW: Do you want to take another look at this?

FtheGFW:
@kirwan  The [513fdf3](https://github.com/facebook/fresco/commit/513fdf32ebb521fc7905b468f72f5f7f36b2af2f) @erikandre pushed already solved the problem I met, and indeed by adding a new status flag to fix the issue is much better 👍 . So I'll just close this PR then.

PR #1781, opened 2017-05-25
Updated Caching docs with up-to-date usage of CacheChoice (instead of ImageType)

Armaxis:
## Motivation

Current [caching documentation](http://frescolib.org/docs/caching.html)'s section on small cache has outdated code example. Had to search myself for what the ImageType was renamed into. 

I updated the code example and corresponding text to refer to CacheChoice properly.

PR #1782, opened 2017-05-26
Docs caching update

Armaxis:
I just noticed that I forgot to replace `ImageType` with `CacheChoice` in next paragraph:

> The pipeline will default to using the same cache for both and ImageType will be ignored.

I already had a [PR for first commit](https://github.com/facebook/fresco/pull/1781) which got approved, but this pr has both commits.


lambdapioneer:
👍  let's merge this as well then

PR #1833, opened 2017-07-20
Support for local content video uri thumbnail loading

IguteChung:
Current local content URI only supports image loading, the thumbnail for video is only supported by file path uri.

### DIFF
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ProducerSequenceFactory.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ProducerSequenceFactory.java
@@ -13,7 +13,6 @@ import java.util.HashMap;
 import java.util.Map;
 
 import android.content.ContentResolver;
-import android.content.Context;
 import android.net.Uri;
 
 import com.facebook.common.internal.Preconditions;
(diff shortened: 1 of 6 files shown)


imagepipeline/src/main/java/com/facebook/imagepipeline/core/ProducerSequenceFactory.java
```diff
@@ -102,6 +107,7 @@ public ProducerSequenceFactory(
       boolean useDownsamplingRatio,
       boolean useBitmapPrepareToDraw,
       boolean partialImageCachingEnabled) {
+    mContentResolver = context.getApplicationContext().getContentResolver();
```
foghina: Could we pass the `ContentResolver` directly instead of the `Context`? Makes it harder to leak things.

imagepipeline/src/main/java/com/facebook/imagepipeline/producers/LocalVideoThumbnailProducer.java
```diff
@@ -113,4 +118,22 @@ private static int calculateKind(ImageRequest imageRequest) {
...
+        null, null, null);
+      try {
+        if (cursor.moveToFirst()) {
+          return cursor.getString(cursor.getColumnIndexOrThrow(MediaStore.Video.Media.DATA));
+        }
+      } finally {
+        cursor.close();
+      }
+    }
+    return "";
```
foghina: Return `null` instead pls, and make sure we handle it correctly down the road (add `@Nullable` etc.)

Issue #675, opened 2015-10-21
HTTP errors are stored in cache


xjy2061:
When connect to a need need authenticate wifi, the response of image request is a html for login, and the html was cached to disk, after that load the image will get from disk cache, but the cache file is not a bitmap file, then the image show nothing, even connect to another wifi does not need authenticate.

michalgr:
Thank you for reporting.

defHLT:
Duplicate of #1413

PR #1845, opened 2017-07-27
#1413, #675: should not put unknown image format data into the disk and encode memory caches

### LINKED ISSUE #1413: Unknown image format exception

zjupure:
The PR aims to fix #1413 , #675. 
fresco will put all the network response with status code 200 into the disk cache and encoded memory cache neither this is image data or not.
When connect to a protal wifi, all network request will rediect to a new site and get a html response including the fresco image request, then image load failed with IllegalArgumentException: unknown image format. Even next time to reload the same url, the image load is still failed due to the incorrect cached data.

I have check the fresco source and find EncodedImage.parseMetaData() will be called in NetworkFetchProducer、LocalFetchProducer、PartialDiskCacheProducer, those producer are executed before DiskCacheWriteProducer、EncodedMemoryCacheProducer. So when the consumer's callback onNewResultImpl() in DiskCacheWriteProducer or EncodedMemoryCacheProducer  is executed and isLast is true, the EncodedImage result will have get the right image format.  If the image format is still unknown, it would be incorrect data and so should not be cached.

### DIFF
--- a/fbcore/src/main/java/com/facebook/common/time/Clock.java
+++ b/fbcore/src/main/java/com/facebook/common/time/Clock.java
@@ -9,9 +9,12 @@
 
 package com.facebook.common.time;
 
+import javax.annotation.concurrent.ThreadSafe;
+
 /**
  * Interface for getting the current time.
  */
+@ThreadSafe
 public interface Clock {
 
   /**
--- a/fbcore/src/main/java/com/facebook/common/time/MonotonicClock.java
+++ b/fbcore/src/main/java/com/facebook/common/time/MonotonicClock.java
@@ -10,10 +10,12 @@
 package com.facebook.common.time;
 
 import com.facebook.common.internal.DoNotStrip;
+import javax.annotation.concurrent.ThreadSafe;
 
 /**
  * A clock that is guaranteed not to go backward.
  */
+@ThreadSafe
 public interface MonotonicClock {
 
   /**


lambdapioneer:
Hey @zjupure, thanks for looking into this. However, it seems that this breaks some unit tests that need fixing :) Also it would be great if you can add new unit tests for the new condition that you've added.

Also I think that the change would also need to be applied to `PartialDiskCacheProducer` (CC @kirwan )

zjupure:
@lambdapioneer broken unit tests have been fixed and one new test are added for the new condition, please review.

As for the `PartialDiskCacheProducer`,  I am not sure the change should be applied to or not. 
line 287 in `PartialDiskCacheProducer`, I see this is place where will write EncodedImage result into partical disk cache and the condition status should have the `IS_PARTICAL_RESULT` flag.
Search the whole source, only 3 places will set the flag `IS_PARTICAL_RESULT`

1. line 165 in OkHttpNetworkFetcher
2. line 137 in PartialDiskCacheProducer
3. line 139 in PartialDiskCacheProducer

case 2 and 3, the consumer is not current `PartialDiskCacheConsumer` but next consumer, so has no effect.
case 1 need the reponse have a header "Content-Range", I can not simluation this situation at present.
and I find the class `PartialDiskCacheConsumer` is added in recent release and is still under experiment.
so I cannot confirm the effect for `PartialDiskCacheConsumer`.

lambdapioneer:
Thank you very much for fixing and adding unit tests! I'll import this for now and let our internal tests run on it.

CC @kirwan who knows about the `PartialDiskCacheProducer` and whether we should include the change there as well.

PR #1848, opened 2017-07-29
Fix issue: should not round null bitmap in RoundedBitmapDrawable

JuniperPhoton:
## Motivation (required)

When setting null bitmap to drawee view which is rounded, it occurs NPE.

## Test Plan (required)

After this fix change, in `DraweeRoundedCornersFragment`,  replace the code in line 88 with:

```
mDraweeRound.getHierarchy().setImage(new BitmapDrawable(getResources(), (Bitmap) null), 1, false);
```

and it should not throw NPE again.

### DIFF
--- a/drawee/src/main/java/com/facebook/drawee/drawable/RoundedBitmapDrawable.java
+++ b/drawee/src/main/java/com/facebook/drawee/drawable/RoundedBitmapDrawable.java
@@ -245,11 +245,11 @@ public class RoundedBitmapDrawable extends BitmapDrawable
   }
 
   /**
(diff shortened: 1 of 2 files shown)


lambdapioneer:
Hey @JuniperPhoton, thank you very much for the pull-request! I'll import it such that our internal tests also run on this.

Can you add unit tests such that this does not regress in future releases? That would be great.

JuniperPhoton:
@lambdapioneer Ok, I will do that.

JuniperPhoton:
@lambdapioneer Unit tests are added.

PR #1849, opened 2017-07-30
Fresco issue 1827: Fix loading of SVG files that do not start with "<svg"

a-d-collins:
## Motivation

Solves the problem of certain SVG files not loading when the SVG file does not begin with "<svg".

## Test Plan

**Manual Test 1: Check loading of example SVG.**
1. Set `ImageFormatSvgFragment.URI_SVG_HALF_TRANSPARENT` equal to `Uri.parse("http://frescolib.org/static/sample-images/fresco_logo_half_transparent.svg");`
2. Run showcase app
3. Ensure that SVG support is turned on
4. (If SVG support was 'off' at the beginning of step 3) Re-run showcase app (<-- This is necessary because the SVG toggle switch does not currently trigger a re-rendering of the image. A follow-up bug has been reported on this.)

**Manual Test 2: Check loading of SVG that starts with "<?xml".**
1. Set `ImageFormatSvgFragment.URI_SVG_HALF_TRANSPARENT` equal to `Uri.parse("https://upload.wikimedia.org/wikipedia/commons/0/02/SVG_logo.svg");`
2. Run showcase app
3. Ensure that SVG support is turned on
4. (If SVG support was 'off' at the beginning of step 3) Re-run showcase app

**Tests added to test.java files**

None.

Note: ImageFormatChecker.determineImageFormat() should probably have a test, though... I'd be happy to write that as part of another pull-request (or add it to this one).

## Most Important Changes

1. (**ImageFormatCheckerUtils.java**) New util that determines whether or not an InputStream contains a given header.
2. (**SvgDecoderExample.java**) When image header starts with "<?xml", search headerBytes for "<svg".

### DIFF
--- a/fbcore/src/main/java/com/facebook/common/memory/PooledByteBufferInputStream.java
+++ b/fbcore/src/main/java/com/facebook/common/memory/PooledByteBufferInputStream.java
@@ -28,6 +28,10 @@ public class PooledByteBufferInputStream extends InputStream {
 
   @VisibleForTesting
   int mOffset; // current offset in the chunk
+
+  // TODO: determine better name for this
+  public static int byteOffset = -1;
+
   @VisibleForTesting
   int mMark; // position of 'mark' if any
 
@@ -39,7 +43,7 @@ public class PooledByteBufferInputStream extends InputStream {
     super();
     Preconditions.checkArgument(!pooledByteBuffer.isClosed());
     mPooledByteBuffer = Preconditions.checkNotNull(pooledByteBuffer);
-    mOffset = 0;
+    mOffset = byteOffset != -1 ? byteOffset : 0;
     mMark = 0;
   }
 
--- a/fbcore/src/main/java/com/facebook/common/memory/PooledByteBufferInputStream.java
+++ b/fbcore/src/main/java/com/facebook/common/memory/PooledByteBufferInputStream.java
@@ -29,11 +29,11 @@ public class PooledByteBufferInputStream extends InputStream {
   @VisibleForTesting
   int mOffset; // current offset in the chunk
 
-  // TODO: determine better name for this
-  public static int byteOffset = -1;
-
   @VisibleForTesting
   int mMark; // position of 'mark' if any
+
+  // TODO: determine better name for this
+  public static int byteOffset;
 
   /**
    * Creates a new inputstream instance over the specific buffer.
@@ -43,7 +43,7 @@ public class PooledByteBufferInputStream extends InputStream {
     super();
     Preconditions.checkArgument(!pooledByteBuffer.isClosed());
     mPooledByteBuffer = Preconditions.checkNotNull(pooledByteBuffer);
-    mOffset = byteOffset != -1 ? byteOffset : 0;
+    mOffset = byteOffset;
     mMark = 0;
   }
 
--- a/fbcore/src/main/java/com/facebook/common/memory/PooledByteBufferInputStream.java
+++ b/fbcore/src/main/java/com/facebook/common/memory/PooledByteBufferInputStream.java
@@ -32,9 +32,6 @@ public class PooledByteBufferInputStream extends InputStream {
   @VisibleForTesting
   int mMark; // position of 'mark' if any
 
-  // TODO: determine better name for this
-  public static int byteOffset;
-
   /**
    * Creates a new inputstream instance over the specific buffer.
    * @param pooledByteBuffer the buffer to read from
@@ -43,7 +40,7 @@ public class PooledByteBufferInputStream extends InputStream {
     super();
     Preconditions.checkArgument(!pooledByteBuffer.isClosed());
     mPooledByteBuffer = Preconditions.checkNotNull(pooledByteBuffer);
-    mOffset = byteOffset;
+    mOffset = 0;
     mMark = 0;
   }
 
--- a/fbcore/src/main/java/com/facebook/common/memory/PooledByteBufferInputStream.java
+++ b/fbcore/src/main/java/com/facebook/common/memory/PooledByteBufferInputStream.java
@@ -28,7 +28,6 @@ public class PooledByteBufferInputStream extends InputStream {
 
   @VisibleForTesting
   int mOffset; // current offset in the chunk
-
   @VisibleForTesting
   int mMark; // position of 'mark' if any
 
--- a/imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormat.java
+++ b/imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormat.java
@@ -11,9 +11,6 @@ package com.facebook.imageformat;
 
 import javax.annotation.Nullable;
 
-import java.io.IOException;
-import java.io.InputStream;
-
 /**
  * Class representing all used image formats.
  */
@@ -35,8 +32,7 @@ public class ImageFormat {
      * @return the image format or null if unknown
      */
     @Nullable
-    ImageFormat determineFormat(byte[] headerBytes, int headerSize, InputStream is)
-      throws IOException;
+    ImageFormat determineFormat(byte[] headerBytes, int headerSize);
   }
 
   // Unknown image format
--- a/imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormat.java
+++ b/imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormat.java
@@ -11,6 +11,9 @@ package com.facebook.imageformat;
 
 import javax.annotation.Nullable;
 
+import java.io.IOException;
+import java.io.InputStream;
+
 /**
  * Class representing all used image formats.
  */
@@ -32,7 +35,8 @@ public class ImageFormat {
      * @return the image format or null if unknown
      */
     @Nullable
-    ImageFormat determineFormat(byte[] headerBytes, int headerSize);
+    ImageFormat determineFormat(byte[] headerBytes, int headerSize, InputStream is)
+      throws IOException;
   }
 
   // Unknown image format
--- a/imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormatChecker.java
+++ b/imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormatChecker.java
@@ -52,13 +52,13 @@ public class ImageFormatChecker {
 
     if (mCustomImageFormatCheckers != null) {
       for (ImageFormat.FormatChecker formatChecker : mCustomImageFormatCheckers) {
-        ImageFormat format = formatChecker.determineFormat(imageHeaderBytes, headerSize);
+        ImageFormat format = formatChecker.determineFormat(imageHeaderBytes, headerSize, is);
         if (format != null && format != ImageFormat.UNKNOWN) {
           return format;
         }
       }
     }
-    ImageFormat format = mDefaultFormatChecker.determineFormat(imageHeaderBytes, headerSize);
+    ImageFormat format = mDefaultFormatChecker.determineFormat(imageHeaderBytes, headerSize, is);
     if (format == null) {
       format = ImageFormat.UNKNOWN;
     }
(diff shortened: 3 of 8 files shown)


fbcore/src/main/java/com/facebook/common/memory/PooledByteBufferInputStream.java
```diff
@@ -28,9 +28,13 @@
 
   @VisibleForTesting
   int mOffset; // current offset in the chunk
+
   @VisibleForTesting
   int mMark; // position of 'mark' if any
 
+  // TODO: determine better name for this
```
lambdapioneer: no need for this TODO here :)

imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormatCheckerUtils.java
```diff
@@ -59,5 +59,31 @@ public static boolean startsWithPattern(
     return true;
   }
 
+  /**
+   * Checks if byteArray interpreted as sequence of bytes contains the pattern.
+   * @param byteArray the byte array to be checked
+   * @param pattern the pattern to check
+   * @return index of beginning of pattern, if found; otherwise -1
+   */
+  public static int containsPattern(
```
lambdapioneer: We can just use `String.contains()` here. No need for writing an O(n^2) find substring.

samples/showcase/src/main/java/com/facebook/fresco/samples/showcase/imageformat/svg/SvgDecoderExample.java
```diff
@@ -50,13 +55,23 @@ public int getHeaderSize() {
...
       }
       if (ImageFormatCheckerUtils.startsWithPattern(headerBytes, HEADER)) {
         return SVG_FORMAT;
       }
+
+      // TODO: (1) Determine if passing the InputStream to this method is the best way to achieve this result.
+      // TODO: (2) Determine if searching the whole array of bytes in the stream is best practice, according to the repository authors.
+      int indexOfPattern = ImageFormatCheckerUtils.containsPattern(ByteStreams.toByteArray(is), HEADER);
+      if (indexOfPattern != -1) {
+        PooledByteBufferInputStream.byteOffset = indexOfPattern;
```
lambdapioneer: If you'd really want the image format checker to have side effects: You probably want to consume the first `indexOfPattern` bytes instead.

imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormat.java
```diff
@@ -32,7 +35,8 @@
      * @return the image format or null if unknown
      */
     @Nullable
-    ImageFormat determineFormat(byte[] headerBytes, int headerSize);
+    ImageFormat determineFormat(byte[] headerBytes, int headerSize, InputStream is)
```
oprisnik: This is quite a breaking API change and further down you're reading the whole input stream in a byte array. Can you not just read the first N bytes instead like we do for all other formats?

a-d-collins: I can read the first N bytes, but I'm not guaranteed to find the header tag of the SVG file if it starts, for example, at the 110th byte and the "max header length" is less than 100 bytes. That said, and after further review of the SVG formatting standards, it's probably best to either (1) leave the original code as it is or (2) detect some of the tags that commonly precede an SVG header tag ("<?xml" and "<!DOCTYPE") and write in a more descriptive Exception message for those scenarios. In both cases, we would ultimately expect the user to provide an image that starts with the "<svg" tag.

Thoughts?

imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormatCheckerUtils.java
a-d-collins: Gotcha. I had originally wanted to follow the style of the startsWithPattern() method above this one, but if you don't think it's necessary here, then that's fine with me. Also, can't we consider the time complexity to be O(n), or at least O(mn), since the length of the pattern is constant? I was sort of following the [String.contains()](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/lang/String.java#String.indexOf%28char%5B%5D%2Cint%2Cint%2Cchar%5B%5D%2Cint%2Cint%2Cint%29) algorithm, because it does basically the same thing.

samples/showcase/src/main/java/com/facebook/fresco/samples/showcase/imageformat/svg/SvgDecoderExample.java
a-d-collins: Yeah. Maybe you're right about that.

imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormat.java
oprisnik: Yeah, I think that something like this would probably work better. Trying to find the actual SVG tag would b

If you know beforehand which images will be SVG, another option would be to just set the SVG decoder as an override that decodes all images and don't do a header check. Or add an image format override (e.g. by adding it to ImageDecodeOptions). Then you could just assume the format for a given image without needing to check the header (and just fail if it's invalid). I'm not sure how hard this would be to implement though.

a-d-collins: Got it. Some of the first part of your message got clipped somehow, but I think I see where you were going with it.

Knowing beforehand which images will be SVG, JPEG, etc. at this point in the code would be nice, but the first question that comes to mind is, "from where do we get that information?" At some point or another, the header of the file will have to be searched to determine the file type. If it's here, great. If it's "before" this point in the workflow, that's fine as well, but until we discover an additional advantage (preferably several others) to having the header-parsing code somewhere other than here, it's probably not worth the effort to move the code.

Maybe the API user could be given the option to provide image types for all of their images. I would be a significant change, but it's certainly interesting to think about.

For now, I'll set up some simple Exception messages like I mentioned in option 2 of my last comment and then I'll leave the merge decision up to you guys.

kirwan:
I'm not keen on breaking the API here.

The bigger problem for me though is that custom format checkers are queried first. As such, every single image you load would run through the entire file seeking `<svg` which assuming most images are JPEGs will be hugely wasteful. And of course you could hit a false positive if you seek through every file in full.

So, I have a couple of thoughts about this.

1) We should alter the order so that the standard formats are checked first. This will remove the possibility of overriding the JPEG decoder but I don't think that was a stated benefit of this API anyway. That way, if a format checker is unavoidably more expensive it will at least only affect the loading of non-standard format images.

2) It may still be imperfect, but how about having the SVG checker check for either `<svg` or `<?xml` at the start of the file. If it finds `<?xml` then it could look through the remainder of the header looking for `<svg` but otherwise bail out at that early stage. Then, editing `getHeaderSize()` to return a higher value (but not ridiculously high, something like 40-50) reduces the chances of false negatives.

What do you think @a-d-collins?

a-d-collins:
@kirwan, I

a-d-collins:
@lambdapioneer, @oprisnik, and @kirwan, I'm done making updates. When you all get the chance, let me know your thoughts.

kirwan:
Thanks @a-d-collins, this looks good to me now.

a-d-collins:
Awesome. Good to know!

fbcore/src/main/java/com/facebook/common/memory/PooledByteBufferInputStream.java
```diff
@@ -28,6 +28,7 @@
 
   @VisibleForTesting
   int mOffset; // current offset in the chunk
+
```
oprisnik: Can you please revert this change in this pull request since this is unrelated.

imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormatChecker.java
```diff
@@ -50,19 +50,20 @@ public ImageFormat determineImageFormat(final InputStream is) throws IOException
     final byte[] imageHeaderBytes = new byte[mMaxHeaderLength];
     final int headerSize = readHeaderFromStream(mMaxHeaderLength, is, imageHeaderBytes);
 
+    ImageFormat format = mDefaultFormatChecker.determineFormat(imageHeaderBytes, headerSize);
```
oprisnik: Unfortunately, our pull request import will squash all of your commits into one, so everything will be in the same changeset. Could you open a separate pull request for changing the order so that the commit title and blame make more sense? Thanks!

imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormatCheckerUtils.java
```diff
@@ -59,5 +59,35 @@ public static boolean startsWithPattern(
...
+  public static int indexOfPattern(
+      final byte[] byteArray,
+      final byte[] pattern) {
+    Preconditions.checkNotNull(byteArray);
+    Preconditions.checkNotNull(pattern);
+    if (pattern.length > byteArray.length) {
+      return -1;
+    }
+
+    for (int i = 0; i < (byteArray.length - pattern.length); i++) {
```
oprisnik: Is the Dalvik compiler smart enough to not always compute byteArray.length - pattern.length? I'm not sure. Might be more efficient to extract the end value.

imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormatCheckerUtils.java
```diff
@@ -59,5 +59,35 @@ public static boolean startsWithPattern(
...
+      if (byteArray[i] == pattern[0]) {
+        for (int j = 1; j < pattern.length; j++) {
+          if (byteArray[i + j] != pattern[j]) {
+            break;
+          }
+          if (j == pattern.length - 1) {
```
oprisnik: I don't think you should do this check in every loop iteration. Maybe you can do something similar as `String.indexOf(...)` here: http://hg.openjdk.java.net/jdk7u/jdk7u2/jdk/file/58ad18490a50/src/share/classes/java/lang/String.java#l1770

samples/showcase/src/main/java/com/facebook/fresco/samples/showcase/imageformat/svg/SvgDecoderExample.java
```diff
@@ -57,6 +58,13 @@ public ImageFormat determineFormat(byte[] headerBytes, int headerSize) {
       if (ImageFormatCheckerUtils.startsWithPattern(headerBytes, HEADER)) {
         return SVG_FORMAT;
       }
+      for (String possibleHeaderTag : POSSIBLE_HEADER_TAGS) {
+        byte[] possibleHeader = ImageFormatCheckerUtils.asciiBytes(possibleHeaderTag);
```
oprisnik: Doing this for each image is unnecessary work. Since you never actually need the String header tag, you can just do: `POSSIBLE_HEADER_TAGS = { ImageFormatCheckerUtils.asciiBytes("<?xml")}`

fbcore/src/main/java/com/facebook/common/memory/PooledByteBufferInputStream.java
a-d-collins: Yes.

imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormatCheckerUtils.java
a-d-collins: I'm not sure either, but I fixed this when I remodeled the whole method to look more like the String.indexOf() method you referenced in your comment below.

imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormatCheckerUtils.java
a-d-collins: Done.

fbcore/src/main/java/com/facebook/common/memory/PooledByteBufferInputStream.java
a-d-collins: Done

imagepipeline-base/src/main/java/com/facebook/imageformat/ImageFormatChecker.java
a-d-collins: Roger that. I will revert these changes and open another pull request that features them.

samples/showcase/src/main/java/com/facebook/fresco/samples/showcase/imageformat/svg/SvgDecoderExample.java
a-d-collins: Makes perfect sense.

PR #1865, opened 2017-08-13
Issue 1850: Restart warning message for showcase SVG support toggle

a-d-collins:
## Motivation

A restart message/warning does not appear when SVG support (svg decoder switch) is changed. A restart is necessary because custom decoders can only be set during Fresco initialization.

## Test Plan

Run showcase app. Check that message comes up when SVG support is turned on/off.


a-d-collins:
Sure thing! You're welcome!

PR #1874, opened 2017-08-22
Switch to OkHttp 3.8.0 for OKHTTP3 users

eldk:
Hello,

Switch to OkHttp 3.8.0 for OKHTTP3 users.

http://square.github.io/okhttp/

https://github.com/square/okhttp/blob/parent-3.8.0/CHANGELOG.md

"OkHttp no longer recovers from TLS handshake failures by attempting a TLSv1 connection"

Thanks,

Eric


oprisnik:
Thanks for the PR!

PR #1892, opened 2017-09-05
webp dependency name changed

simonzhexu:
Correct webp doc to include the correct dependency library name.


lambdapioneer:
Good catch and thanks for writing this pull-request! I'll land it now

simonzhexu:
Thanks Daniel!

PR #1902, opened 2017-09-13
Add rounding support for RTL languages

### LINKED ISSUE #1898: Add rounding support for RTL languages to GenericDraweeView
Can we add rounding support for RTL languages to GenericDraweeView?

Need to add:
```
fresco:roundTopStart
fresco:roundTopEnd
fresco:roundBottomStart
fresco:roundBottomEnd
```

Android Studio already recommends using start/end instead of left/right in support of RTL languages (e.g., android:paddingLeft -> android:paddingStart)

I would be happy to help contribute too.

thanks,
Adam

adamc01:
Fixes #1898.

 Screenshots of fixed version:

screenshot_20170913-101639
screenshot_20170913-101257
screenshot_20170913-101102
screenshot_20170913-100906


drawee/src/main/java/com/facebook/drawee/generic/GenericDraweeHierarchyInflater.java
```diff
@@ -190,6 +212,20 @@ public static GenericDraweeHierarchyBuilder updateBuilder(
...
         gdhAttrs.recycle();
+
+        if (android.os.Build.VERSION.SDK_INT >= 17 && context.getResources().getConfiguration()
+            .getLayoutDirection() == View.LAYOUT_DIRECTION_RTL) {
+          roundTopLeft = roundTopLeft && roundTopEnd;
+          roundTopRight = roundTopRight && roundTopStart;
+          roundBottomRight = roundBottomRight && roundBottomStart;
+          roundBottomLeft = roundBottomLeft && roundBottomEnd;
+        } else {
+          roundTopLeft = roundTopLeft && roundTopStart;
```
erikandre: Thanks for the pull request @adamc01 !

Overall it looks great but I think we have an issue where when it comes to combining roundXstart/end with roundXleft/right.

If you compare this to how setting margins or padding for RTL is done:

**Set left margin that should become right margin for RTL locales**
android:layout_marginStart="10dp"
android:layout_marginLeft="10dp"

While here its reversed:

roundTopLeft = roundTopLeft && roundTopEnd;

In the case of RTL layout direction on Android 4.2+, the top left corner should be rounded if

roundTopEnd==true && roundTopRight is either true or not set

Let me know if this is clear and if it makes sense to you. Thanks!

adamc01: On API 17+ force RTL mode (make sure `android:supportsRtl="true"` is set in the AndroidManifest) and if you do the following:
```
      <com.facebook.drawee.view.SimpleDraweeView
          android:id="@+id/drawee_some_rtl"
          style="@style/DraweeRoundedCorners.Drawee"
          app:roundedCornerRadius="16dp"
          android:layout_marginStart="50dp"
          android:layout_marginLeft="50dp"
          />
```

then Start does indeed override Left. However if you do:
```
      <com.facebook.drawee.view.SimpleDraweeView
          android:id="@+id/drawee_some_rtl"
          style="@style/DraweeRoundedCorners.Drawee"
          app:roundedCornerRadius="16dp"
          android:paddingStart="50dp"
          android:paddingLeft="50dp"
          />
```
Start does not override Left and you can see both sides have padding.

My implementation is consistent with this (although this may be a bug in AOSP: https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/core/java/android/view/View.java#17131 ?) and frankly the developer should not be mixing and matching, but if they do, then it's WYSIWYG. You can set a Start, and also set a Left and Left will always set on the left side regardless of RTL/LTR orientation, and Start will adjust accordingly. That seems more intuitive than Google's implementation on margin where they override it.


> In the case of RTL layout direction on Android 4.2+, the top left corner should be rounded if roundTopEnd==true && roundTopRight is either true or not set

I did the following:
```
      <com.facebook.drawee.view.SimpleDraweeView
          android:id="@+id/drawee_some_rtl"
          style="@style/DraweeRoundedCorners.Drawee"
          app:roundedCornerRadius="16dp"
          app:roundTopEnd="true"
          app:roundTopRight="false"
          />
```
and the top left corner was rounded in RTL mode for me. Did you see something different?

erikandre: Hi @adamc01, sorry for taking a while to get back to you.

OK, that sounds reasonable and after checking the source for View.java I agree that this is consistent with the platform behaviour (as confusing as that may be).

Great work! I'll get this merged now

oprisnik:
Thanks for the PR, it will be included in our next release :)

PR #1989, opened 2017-12-13
Remove SoLoaderShim, use SoLoader

foghina:
## Motivation (required)

SoLoader is more stable on older versions of Android. Switching to it should fix recurring issues that we get such as `UnsatisfiedLinkError` and so on.

## Test Plan (required)

CI + run showcase app.


foghina:
Rebased and using the latest SoLoader.

PR #1990, opened 2017-12-13
Fix all build.gradle files to use new deps declarations

foghina:
## Motivation (required)

We were getting a lot of deprecation warnings when building with gradle after the update to 4.1. This PR switches all `provided` to `compileOnly` and `compile` to `implementation` / `api`, depending on the use case.

## Test Plan (required)

CI + run showcase app.

### DIFF
--- a/imagepipeline-base-test/build.gradle
+++ b/imagepipeline-base-test/build.gradle
@@ -5,11 +5,12 @@ project.group = GROUP
 version = VERSION_NAME
 
 dependencies {
-    compile "com.android.support:support-core-utils:${SUPPORT_LIB_VERSION}"
-    provided "com.google.code.findbugs:jsr305:${JSR_305_VERSION}"
-    compile "com.parse.bolts:bolts-tasks:${BOLTS_ANDROID_VERSION}"
-    provided "javax.annotation:javax.annotation-api:${ANNOTATION_API_VERSION}"
-    compile project(':fbcore')
+    compileOnly "com.google.code.findbugs:jsr305:${JSR_305_VERSION}"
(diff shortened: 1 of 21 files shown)


imagepipeline-base-test/build.gradle
```diff
@@ -36,7 +37,7 @@ task jarTest (type: Jar) {
 }
 
 configurations {
-    testOutput.extendsFrom (testCompile)
+    testOutput.extendsFrom (testimplementation)
```
oprisnik: Don't you want `testImplementation`?

imagepipeline-base/build.gradle
```diff
@@ -5,28 +5,30 @@ project.group = GROUP
...
+    compileOnly "com.google.code.findbugs:jsr305:${JSR_305_VERSION}"
+    compileOnly "javax.annotation:javax.annotation-api:${ANNOTATION_API_VERSION}"
 
-    testCompile "junit:junit:${JUNIT_VERSION}"
-    testCompile "org.easytesting:fest-assert-core:${FEST_ASSERT_CORE_VERSION}"
-    testCompile "org.mockito:mockito-core:${MOCKITO_CORE_VERSION}"
-    testCompile("org.powermock:powermock-api-mockito:${POWERMOCK_VERSION}") {
+    implementation "com.parse.bolts:bolts-tasks:${BOLTS_ANDROID_VERSION}"
+    implementation project(':fbcore')
+    implementation 'com.facebook.soloader:soloader:0.1.0'
```
oprisnik: can you also do the same versioning as we do for the others? Also, should this be in this diff?

foghina:
Whoops, some of the changes from #1989 butted their way in here. Removed them and fixed the `testimplementation` thing. Not sure how the build was passing with that though.

Issue #2007, opened 2017-12-26
CacheKey doesn't contains Postprocessor when submit a ImageRequest


jungletian:
### Description

When submit a ImageRequest with PostProcessor, BitmapMemoryCacheProducer$produceResults() should use `getPostprocessedBitmapCacheKey` method. Otherwise the cached bitmap key is different from the `mCacheKey` of PipelineDraweeController.class

### Solution

use `getPostprocessedBitmapCacheKey` replace `getBitmapCacheKey` in BitmapMemoryCacheProducer$produceResults()

zmroczek:
This issue isn't valid since the logic for caching post-processed images is included in PostProcessorProducer (more info in #2008).

zmroczek:
This PR changes the current logic of caching post-processed images. This issue isn't valid because in order to cache the post-processor results you need to implement the getPostprocessorCacheKey() method (see docs http://frescolib.org/docs/modifying-image.html#caching-post-processor-results) which is used in PostprocessorProducer.

For postprocessors that we decide not to require caching the image to be processed is still taken from cache with a cache key returned from getBitmapCacheKey() method.

PR #2008, opened 2017-12-26
fix issue #2007 use getPostprocessedBitmapCacheKey replace getBitmapC…

jungletian:
fixed issue #2007

this should use `getPostprocessedBitmapCacheKey` when imageRequest.getPostProcessor() is not null.


oprisnik:
Thanks for the pull request @jungletian. Can you please sign the CLA?

jungletian:
@oprisnik done

zmroczek:
This PR changes the current logic of caching post-processed images. This issue isn't valid because in order to cache the post-processor results you need to implement the `getPostprocessorCacheKey()` method (see docs http://frescolib.org/docs/modifying-image.html#caching-post-processor-results) which is used in `PostprocessorProducer`.

For postprocessors that we decide not to require caching the image to be processed is still taken from cache with a cache key returned from `getBitmapCacheKey()` method.

Issue #2006, opened 2017-12-26
the wasImmediate is always false.


jungletian:
https://github.com/facebook/fresco/blob/fb0aa4fd16d37d340d83a6d4455a2ec533ed0cfa/drawee/src/main/java/com/facebook/drawee/controller/AbstractDraweeController.java#L453

zmroczek:
In the current code `wasImmediate` is not always `false`. Please take a look on tests in `AbstractDraweeControllerTest` where you can find scenarios when it is `true`.

The PR you sent is breaking the tests for scenarios where `wasImmediate` should be `false`, it seems like after your change it always returns `true` which is not expected.

PR #2010, opened 2017-12-27
fix issue #2006: the wasImmediate should be obtained when dataSource …

jungletian:
fix issue #2006 

cc


zmroczek:
I'm closing this PR since this is breaking the correct behavior of `wasImmediate` (more info in #2006).

PR #2018, opened 2018-01-09
Do not treat response as partial with Content-Range if range includes…

bschnack:
Do not treat the response as partial when "Content-Range" header is set if range includes the entire message.

## Motivation (required)

If the header "Content-Range" is ever passed in from a network request, the image will not be saved to cache. What I have changed is that, if "Content-Range" includes the entire content range for the image, then do not treat as a partial image.

## Test Plan (required)

I have tested this on my personal repo by trying to fetch an image with "Content-Range" set, but I have not created any unit tests.


kirwan:
I think we should only make this change if the range also begins at byte 0. If you want to make that change here then we can merge this.

I'm curious about cases where the response include the `Content-Range` header despite returning the entire response.

Can you share a URL to an image which needs this change?

bschnack:
Hello Kirwan,

I have updated the PR to check both to and from. Unfortunately, I do not have an unauthenticated URL to an image that has Content-Range set but is returning the entire response. You could add the header if you add a breakpoint (I discovered this header was the issue by removing the header at a breakpoint).

Issue #2041, opened 2018-02-02
okhttp3 may not build after Fresco sources update


eldk:
### Description

Build from sources complains and outputs errors (even after clean project) :  missing okhttp3 after this commit (switch from compile to implementation) : https://github.com/facebook/fresco/commit/b83a71b9d27cd3d95ef59cdaa291d89b6a09c27a#diff-68af0e80447eb282f6a40fa64a9b16f7 - https://github.com/facebook/fresco/blob/06fe3719a3a3c659438b6c25a8008945e67156a0/imagepipeline-backends/imagepipeline-okhttp3/build.gradle#L7

### Workaround
switch back to compile : 
`compile "com.squareup.okhttp3:okhttp:${OKHTTP3_VERSION}" `

Clean and build, then switch back to : 
 `implementation "com.squareup.okhttp3:okhttp:${OKHTTP3_VERSION}"`

### Additional Information
Not a bug - Workaround 
* Fresco version: up to https://github.com/facebook/fresco/commit/b83a71b9d27cd3d95ef59cdaa291d89b6a09c27a#diff-68af0e80447eb282f6a40fa64a9b16f7
* Gradle version : 4.1
* Android studio version : 3.0.1

eldk:
Hello,

Just test again, when debug build with use of 'implementation', it always fails.

oprisnik:
Thanks for the PR @eldk !

PR #2042, opened 2018-02-02
Gradle : switch from implementation to api. #2041

eldk:
Hello,

solves https://github.com/facebook/fresco/issues/2041
explanation : https://stackoverflow.com/questions/44413952/gradle-implementation-vs-api-configuration/44419574#44419574

Thanks,

Eric

### DIFF
--- a/imagepipeline-backends/imagepipeline-okhttp/build.gradle
+++ b/imagepipeline-backends/imagepipeline-okhttp/build.gradle
@@ -4,7 +4,7 @@ project.group = GROUP
 version = VERSION_NAME
 
 dependencies {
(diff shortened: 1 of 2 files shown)


lambdapioneer:
Hi @eldk, thanks for the pull-request :) I'm going to check with @foghina if this change was intended this way.

It's coincidence that the showcase builds fine because the `implementation "com.facebook.stetho:stetho-okhttp3:${STETHO_VERSION}"` dependency leaks OkHttp3 as a `compile` dependency.

I'm happy to change it back to API as most people want to access `OkHttp`/`OkHttp3` in their application any way. However, it would also be good to have them make it explicit in their own gradle file :)

Issue #2040, opened 2018-02-02
Debug build from sources fails with error : use of undeclared identifier 'memset'


eldk:
### Description

In Android Studio, up to date, with ndk 16.1.4479499, and Fresco lib sources from master branch, to last commit (yesterday), build fails with : 
```
./jpeg/jpeg_codec.cpp:180:3: error: use of undeclared identifier 'memset'; did you mean 'wmemset'?
  memset(&dinfo, 0, sizeof(struct jpeg_decompress_struct));
```
(...) and few more errors (...)
```
1: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':imagepipeline:ndk_build_imagepipeline'.
```
```
2: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':static-webp:ndk_build_static-webp'.
```
### Reproduction

Try to build again

* Fresco version: master to https://github.com/facebook/fresco/commit/06fe3719a3a3c659438b6c25a8008945e67156a0
* Platform version: debug build for api 24 - Android 6.0

lambdapioneer:
Thanks for the bug report @eldk!

Looks like both copies of `jpeg_codec` are missing `#include <string.h>`. I'd guess that previous NDK versions just did not care enough about this.

Can you check if adding those lines to the file might solve the problem? I don't have that NDK version ready at the moment - but otherwise I'll have a look on Monday :)

eldk:
Hello,
I'm switching back to 15c now.
I will test if it should be a temporary workaround.
Thanks,
Eric

eldk:
Thanks @lambdapioneer

PR #2043, opened 2018-02-02
Debug build from sources fails with error : undeclared identifier memset

eldk:
ndk : 16.1.4479499
android-studio : 3.01
solves : https://github.com/facebook/fresco/issues/2040

Thanks @lambdapioneer


lambdapioneer:
Awesome! Thank you very much :) I will just check if it also builds fine with older NDK tool chains

PR #2044, opened 2018-02-03
Upgrade static-webp tests librairies

eldk:
Upgrade tests librairies.


lambdapioneer:
Looks good to me :) Was there any specific issue that arouse from having the older versions?

eldk:
Hello,

It's to avoid conflict when multidex is enable as ''com.android.support.test:runner:1.0.1' is used in other Fresco modules and should be used in non Fresco ones.

`samples/showcase/build.gradle:    androidTestImplementation 'com.android.support.test:runner:1.0.1`

Thanks,
Eric

PR #2068, opened 2018-03-14
Default request timeout

jnawaz:
## Motivation (required)
The motivation for this change comes from a problem in an app I'm working on, under very bad connectivity there was an issue where timeout meant the image wasn't being downloaded. After lots of testing and consideration to swap to a regular ImageView and download a Bitmap, I investigated further and found under the worst conditions it was taking 20 seconds to download the image.

What existing problem does the pull request solve?
It solves a timeout issue when downloading images over a weak or poor connection.

## Test Plan (required)
Unit test added to confirm connection timeout is set.

### DIFF
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipelineConfig.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipelineConfig.java
@@ -80,6 +80,7 @@ public class ImagePipelineConfig {
   private final DiskCacheConfig mMainDiskCacheConfig;
   private final MemoryTrimmableRegistry mMemoryTrimmableRegistry;
   private final NetworkFetcher mNetworkFetcher;
+  private int mHttpNetworkTimeout = -1;
   @Nullable private final PlatformBitmapFactory mPlatformBitmapFactory;
   private final PoolFactory mPoolFactory;
   private final ProgressiveJpegConfig mProgressiveJpegConfig;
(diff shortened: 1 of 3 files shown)


imagepipeline/src/main/java/com/facebook/imagepipeline/producers/HttpUrlConnectionNetworkFetcher.java
```diff
@@ -34,6 +34,8 @@
   public static final int HTTP_TEMPORARY_REDIRECT = 307;
   public static final int HTTP_PERMANENT_REDIRECT = 308;
 
+  public static final int HTTP_DEFAULT_TIMEOUT = 30000;
```
oprisnik: Can you make this value configurable please? Either by adding a second constructor and/or adding a setter for this.
This allows people to customize the timeout depending on their needs.

jnawaz: Of course, it makes sense to make it configurable to the users personal need.

jnawaz:
HttpConnectionTimeout is now configurable but still defaults to 30 seconds if the user doesn't want to specify a timeout, thereby still fixing the original issue but allowing custom timeout.

imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipelineConfig.java
```diff
@@ -80,6 +80,7 @@
   private final DiskCacheConfig mMainDiskCacheConfig;
   private final MemoryTrimmableRegistry mMemoryTrimmableRegistry;
   private final NetworkFetcher mNetworkFetcher;
+  private int mHttpNetworkTimeout = -1;
```
oprisnik: can you make this final please?

imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipelineConfig.java
```diff
@@ -390,6 +396,11 @@ public Builder setCacheKeyFactory(CacheKeyFactory cacheKeyFactory) {
       return this;
     }
 
+    public Builder setHttpConnectionTimeout(int httpConnectionTimeout) {
```
oprisnik: It would be better to include the unit at the end for all fields (e.g. `httpConnectionTimeoutMs`) so that you know what number you'd have to set.

imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipelineConfig.java
jnawaz: Marked as final and modified code in ImagePipelineConfig to handle change to final declaration.

imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipelineConfig.java
jnawaz: Added Ms for milliseconds as suggested.

jnawaz:
Changes made as requested.

PR #2082, opened 2018-03-30
Feature/ci

gengjiawen:
## Motivation (required)

* current circleci build has no instrumentation build, so I create an avd to do the test.
* current build has no test against the proguard rules, so I enable minify in sample project main flavor.

Q and A:
* why not some new avd like sdk 23 or higher
The latest armeabi avd (which I find is 23) has some lock screen issue `Waited for the root of the view hierarchy to have window focus and not request layout for 10 seconds.` ([full log](https://circleci.com/gh/gengjiawen/fresco/38#tests/containers/0)).And x86 arch can't use inside docker.
* Why ci cache added, is that really necessary
When I made this pr, lots of time I have to fight 302 and 404 in bintray.com, so I think it could save us time.
* Any further improvement
We can maintain our docker and get rid of the one circleci offered. So we no longer needed to download the avd and ndk. 

## Test Plan (required)

pass ci


circle.yml
```diff
@@ -17,13 +30,38 @@ jobs:
...
+            sdkmanager "system-images;android-19;default;armeabi-v7a"
+            echo "no" | avdmanager create avd -n test -k "system-images;android-19;default;armeabi-v7a"
+      - run:
+          name: Launch emulator
+          command: emulator -avd test -no-audio -no-window
+          background: true
+      - run:
+          name: Wait for emulator
+          command: |
+            wget --quiet --output-document=android-wait-for-emulator https://raw.githubusercontent.com/travis-ci/travis-cookbooks/master/community-cookbooks/android-sdk/files/default/android-wait-for-emulator
```
foghina: I'm a bit wary of downloading something and running it. Could we instead add a snapshot of this script to our repo?

samples/showcase/build.gradle
```diff
@@ -21,15 +21,25 @@ android {
...
         }
     }
+
     flavorDimensions "main"
     productFlavors {
-        main
+        main {
+            buildTypes {
+                debug {
+                    minifyEnabled true
```
foghina: Does this mean we're going to minify all debug builds if instant run is turned off? Could we instead have a separate "instrumentation" target?

circle.yml
gengjiawen: Yes, I think in the long run we should we use fresco own docker instead of the official one. In this way we can also remove download ndk part.

samples/showcase/build.gradle
gengjiawen: It's only in the main flavor, the `instantRunConfig` flavor is not affected. If main flavor is used a lot, I think we can create a new flavor just for instrumentation test.

circle.yml
foghina: Sure, I might work on that at some point. In the meantime we should still add this script to our repo and call it locally. Do you think you can do that?

samples/showcase/build.gradle
foghina: I think main flavour is used a bunch, since instantRun is still kinda buggy. Do you mind adding a separate buildType?

gengjiawen:
I tested avd 22, looks like have memory issue, https://circleci.com/gh/gengjiawen/fresco/46.

foghina:
@gengjiawen that looks like it just failed due to a transient issue while running plain unit tests, no? It doesn't look to me like it even tried to run instrumentation tests.

samples/showcase/build.gradle
gengjiawen: I add a flavor called internal to do instrumentation test.

gengjiawen:
@foghina yes, that's werid. I suspect it's the memory issue with circle ci. But when you downgrade the avd version, it works fine.

gengjiawen:
To be honest, I have been using `system-images;android-24;default;armeabi-v7a` for my gitlab project, which works fine. I want to report it to circle ci, but this is which I find a little difficulty for now.

foghina:
Thanks again for doing this!

lambdapioneer:
This is amazing 🎉 thanks a lot!

PR #2110, opened 2018-05-09
[fresco] add dropBitmapAnimationBackend for webp list

woyunowuyuda:
## Motivation (required)
Fresco is very powerful, but recently we discovered the following problems when used:
In a list of all dynamic WEBP diagrams (approximately 16 elements on a single screen), WEBP is loaded using SimpleDraweeView and setAutoPlayAnimations(true) , there will be a noticeable stuck in either the WEBP play or the list slide. We followed up the stuck and found:
BitmapAnimationBackend renders and draws frames on the UI thread when the cache does not exist. This takes too much time in the WEBP list, resulting in a stuck. This submission is to solve.

## Test Plan (required)
In a list which all play WEBP.

### DIFF
--- a/animated-drawable/src/main/java/com/facebook/fresco/animation/bitmap/DropBitmapAnimationBackend.java
+++ b/animated-drawable/src/main/java/com/facebook/fresco/animation/bitmap/DropBitmapAnimationBackend.java
@@ -1,0 +1,309 @@
+/*
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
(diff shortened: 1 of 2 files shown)


oprisnik:
Thanks for the pull request! This change duplicates a lot of code and I think there should be an easier way to prevent this from happening (e.g. by adding a flag to existing backends to just drop frames when not cached). 

I hope to have some time to take a closer look next week. In the meantime, you could explore simpler approaches that do not duplicate a lot of the existing functionality.

PR #2132, opened 2018-06-09
[ROUND] fix webp add round draw out of bound

woyunowuyuda:
## Motivation (required)
When I add a OVER_COLOR Round to webP Drawable,the overColor draw out of bound and start blinking.

### Before
before
A close look:
before1

### After
after
A close look:
after1

## Test Plan (required)
Add roundedCornerRadius to SimpleDraweeView which plays webP.


oprisnik:
Thanks for the pull request @woyunowuyuda. I'll import for internal testing.

PR #2140, opened 2018-06-21
Fix 2087 round gif

hetovar:
## Motivation

Making rounding GIF would make the application more appealing to users as well as adding some more functionality.

## Test Plan 
Added some test in order to show that the drawable is able to be drawn with either the property or without it.


erikandre:
Closing this as we are not able to accept the contribution as-is and no updates has been made since July.

PR #2143, opened 2018-06-25
[FIX] fix below androidN, the preActivity's DraweeView can't detach when startActivity

woyunowuyuda:
## Motivation (required)

Below Android N, the preActivity's DraweeView can't detach when startActivity. When we startActivity, the PhoneWindow#DecorView invoke onVisibilityChanged(), but at that time getVisiblity() is not equal with the changing visibility. For example, when we startActivity, visibility is going to INVISIBLE but getVisibility() in maybeOverrideVisibilityHandling() is returning VISIBLE.

## Test Plan (required)

Use a phone which API is below N, startActivity to another Acitivity and watch the preActivity's DraweeView has detached or not.

### DIFF
--- a/drawee/src/main/java/com/facebook/drawee/view/DraweeView.java
+++ b/drawee/src/main/java/com/facebook/drawee/view/DraweeView.java
@@ -283,14 +283,18 @@ public class DraweeView<DH extends DraweeHierarchy> extends ImageView {
       View changedView,
       int visibility) {
     super.onVisibilityChanged(changedView, visibility);
-    maybeOverrideVisibilityHandling();
+    maybeOverrideVisibilityHandling(visibility);
   }
 
   private void maybeOverrideVisibilityHandling() {
+    maybeOverrideVisibilityHandling(getVisibility());
+  }
+
+  private void maybeOverrideVisibilityHandling(int visibility) {
     if (mLegacyVisibilityHandlingEnabled)  {
       Drawable drawable = getDrawable();
       if (drawable != null) {
-        drawable.setVisible(getVisibility() == VISIBLE, false);
+        drawable.setVisible(visibility == VISIBLE, false);
       }
     }
   }


zmroczek:
Hi @woyunowuyuda

Thanks for the PR! 

It seems that the code you changed for getting visibility wouldn't be reached for android versions below N ([because of `mLegacyVisibilityHandlingEnabled` check](https://github.com/woyunowuyuda/fresco/blob/3fcab25f93af6f94ce7731616184978b5b5480d9/drawee/src/main/java/com/facebook/drawee/view/DraweeView.java#L94)), so it only affects versions above. Could you please clarify if this fixes the issue for Android versions below N or above N?

woyunowuyuda:
@zmroczek Hi, because mLegacyVisibilityHandlingEnabled is the logic which already exist, so I didn't change it. Before this PR, below android N, the AbstractDraweeController would not detach  even  mLegacyVisibilityHandlingEnabled is true because of the reason I have mentioned before. This PR fixed this issue so if you set mLegacyVisibilityHandlingEnabled true, the AbstractDraweeController can correctly detach. By the way, how to deeplink a code line like you?

zmroczek:
Do you mean you're using `setLegacyVisibilityHandlingEnabled(true)` method for versions below Android N? The code in [`maybeOverrideVisibilityHandling()`](https://github.com/woyunowuyuda/fresco/blob/3fcab25f93af6f94ce7731616184978b5b5480d9/drawee/src/main/java/com/facebook/drawee/view/DraweeView.java#L295) method should be only executed for Android versions above N, since this is a fallback only for those versions (we will update the setter).

Otherwise, I don't exactly understand what is the issue you're trying to fix for versions below N - could you please provide more steps to reproduce in the showcase app or provide a some sample app with behavior that fails without this PR?

And to link a specific line in the code you can click on the line number and then copy the link from the browser that will end up with e.g. _#L295_ which is the line number, then link it to the text :)

woyunowuyuda:
@zmroczek The example can be as follows:
ActivityA(jump, no finish)   ------>    ActivityB
The ActivityA has a SimpleDraweeView, then jump to AcitivtyB( a empty activity).
Then, you can debug at [AbstractDraweeController#onDetach](https://github.com/facebook/fresco/blob/b67a2d5a4fdccaeb8d2fcc82377b5d4ea76aa25a/drawee/src/main/java/com/facebook/drawee/controller/AbstractDraweeController.java#L366)
Above android N, this can be reached.
Below android N, it will never invoke, which means [mFetchedImage](https://github.com/facebook/fresco/blob/b67a2d5a4fdccaeb8d2fcc82377b5d4ea76aa25a/drawee/src/main/java/com/facebook/drawee/controller/AbstractDraweeController.java#L91) can not release.

oprisnik:
Can you create a sample project that explains the issue and the fix with screenshots please?

erikandre:
Closing this as we cannot verify the issue that it reportedly fixes and we also have some concerns regarding if it's safe or not.

Please comment with more details (preferrably with a complete example project which can be used to demonstrate the issue and the fix) and with that we may be able to move forward. Thanks!

PR #2145, opened 2018-06-29
[ROUND] add BITMAP_ROUND to webp animamation

woyunowuyuda:
## Motivation (required)

I found out that AnimationDrawable2 which used to play webP animation does not support Bitmap round. Because the real draw is under BitmapAnimationBackend class, so I can not use a delegate Drawable or subclass of AnimationDrawable2 to fix it like RoundedBitmapDrawable.I know this is not the beat idea which will make AnimationDrawable2 take too much responsibility, but I can't find a better achievement currently.

### before
beforewebpcorner
### after
afterwebpcover

## Test Plan (required)

Use bitmap round method in animation webP.


oprisnik:
I'm going to close this one as well. Please feel free to submit a new pull request :)

PR #2177, opened 2018-08-08
Libjpeg update to 1.5.3

dryganets:
## Motivation
The libjpeg-turbo version currently distributed with Fresco has major security issues.
For instance this one: https://nvd.nist.gov/vuln/detail/CVE-2016-3616
There are more issues marked as fixed, this one has a HIGH severity rating.

## Test Plan (required)
Run unit tests.


static-webp/src/main/jni/third-party/libjpeg-turbo-1.5.3/config.h
```diff
@@ -110,18 +110,19 @@
...
 
 /* Define to the version of this package. */
-#define PACKAGE_VERSION "1.5.0"
+#define PACKAGE_VERSION "1.5.3"
 
 /* Define if your (broken) compiler shifts signed values as if they were
    unsigned. */
 /* #undef RIGHT_SHIFT_IS_UNSIGNED */
-
+// TODO: double check if it was done on purpose. jconfig.h has different defines for 64 and 32 bits
```
dryganets: I wonder if it is intentionally was changed to 1.

The same thing is defined as
```
#ifdef __LP64__
#  define SIZEOF_SIZE_T 8
#else
#  define SIZEOF_SIZE_T 4
#endif
```

dryganets:
I added the SIZEOF change to the PR as I think it is a right thing to do.

lambdapioneer:
@dryganets , I agree that the previous code (with value 1) was wrong. Thanks for changing it to the correct version that assigns to 4 / 8 bytes depending on the architecture!

PR #2179, opened 2018-08-09
libwebp updated to 1.0.0

dryganets:
## Motivation (required)
Our security check tool found the CVE problem in libwebp.
I think it was a false positive as the related problem was fixed on 0.5.2.

It is up to you to decide if it worth it to update to 1.0.0.

## Test Plan (required)
Make sure that unit tests pass.


defHLT:
Thank you this and other pull requests!

What tool do you use?

dryganets:
This is an internal tool, for us, it looks more like a process - whenever you add a new library in the product you have to pass through the code scan. I think the vendor performing code scan has the tool or database of the CVE's related to the particular library/version.

PR #2189, opened 2018-08-20
Add support for windows ico image format

zjupure:
## Motivation

There is some case we need load .ico format image to SimpleDraweeView and BitmapFactory support decode this format on Android System but fresco not support. Though we can write our own CustomDecoder and config it into fresco, we expect the fresco could support it natively. This can be achieved by only add a little code in `DefaultImageFormatChecker`. 

## Test Plan

http://static.iqiyi.com/ext/common/ffb4f4483feb4c78b9cb74b7eee8d5d3.ico  This ico file can be used for test. It works well on my devices.


oprisnik:
Hi @zjupure. Thanks for the pull request. You don't need a custom decoder or anything else? It just works if you properly handle headers so that the image is passed on to the system?

lambdapioneer:
Looks legit: https://github.com/aosp-mirror/platform_frameworks_base/blob/master/core/jni/android/graphics/BitmapFactory.cpp#L64-L66

Issue #2173, opened 2018-08-05
"ProgressiveDecoder: unknown image format" for some Android URIs


bamsbamx:
### Description

Trying to load the URI identifier for installed applications icons fails for some apps (system and non-system) and throws some logs:
`D/skia: --- SkAndroidCodec::NewFromStream returned null
W/unknown:ProgressiveDecoder: unknown image format, {uri: android.resource://com.sonymobile.synchub/2131427328, firstEncodedBytes: 03000800840100000100, length: 388}
D/skia: --- SkAndroidCodec::NewFromStream returned null
W/unknown:ProgressiveDecoder: unknown image format, {uri: android.resource://com.google.android.calendar/2130903042, firstEncodedBytes: 03000800480200000100, length: 584}
D/skia: --- SkAndroidCodec::NewFromStream returned null
W/unknown:ProgressiveDecoder: unknown image format, {uri: android.resource://com.android.chrome/2131231050, firstEncodedBytes: 03000800540100000100, length: 340}
D/skia: --- SkAndroidCodec::NewFromStream returned null
W/unknown:ProgressiveDecoder: unknown image format, {uri: android.resource://com.sonyericsson.organizer/2130903043, firstEncodedBytes: 03000800C80100000100, length: 456}
D/skia: --- SkAndroidCodec::NewFromStream returned null
D/skia: --- SkAndroidCodec::NewFromStream returned null
W/unknown:ProgressiveDecoder: unknown image format, {uri: android.resource://com.google.android.apps.docs.editors.docs/2130838355, firstEncodedBytes: 030008008C0200000100, length: 652}
D/skia: --- SkAndroidCodec::NewFromStream returned null
D/skia: --- SkAndroidCodec::NewFromStream returned null
W/unknown:ProgressiveDecoder: unknown image format, {uri: android.resource://com.google.android.apps.docs/2130837900, firstEncodedBytes: 03000800280200000100, length: 552}
D/skia: --- SkAndroidCodec::NewFromStream returned null
D/skia: --- SkAndroidCodec::NewFromStream returned null
W/unknown:ProgressiveDecoder: unknown image format, {uri: android.resource://com.sonymobile.email/2130903040, firstEncodedBytes: 03000800280200000100, length: 552}
D/skia: --- SkAndroidCodec::NewFromStream returned null
D/skia: --- SkAndroidCodec::NewFromStream returned null
[...]

Issue #1463, opened 2016-09-08
Should support VectorDrawable resource id


johnzheng00:
<vector> tag need support. v21 and AppCompat already support v9 and up api level.

kirwan:
Where do you want to use VectorDrawable resource IDs?

Right now, Fresco will only load bitmaps for the primary image but for placeholders, backgrounds, overlays, loading drawables and failures they should work.

If you could give a small code snippet we can double check what your issue is.

johnzheng00:
<vector xmlns:android="http://schemas.android.com/apk/res/android"
        android:width="24dp"
        android:height="24dp"
        android:viewportHeight="24.0"
        android:viewportWidth="24.0">
    <path
        android:fillColor="#cccccc"
        android:pathData="M8.59,16.34l4.58,-4.59 -4.58,-4.59L10,5.75l6,6 -6,6z"/>
</vector>
 This is a vector image  on    res/drawable/ic_arrow.xml
So i need load the vector image with Fresco.   
SimpleDraweeView.setImageUri("res://package/id")

johnzheng00:
`<vector xmlns:android="http://schemas.android.com/apk/res/android"
        android:width="24dp"
        android:height="24dp"
        android:viewportHeight="24.0"
        android:viewportWidth="24.0">
    <path
        android:fillColor="#cccccc"
        android:pathData="M8.59,16.34l4.58,-4.59 -4.58,-4.59L10,5.75l6,6 -6,6z"/>
</vector>
`

kirwan:
`DraweeView` doesn't support VectorDrawables as the main image, only bitmaps. I think you would need to simply use an `ImageView` in this case.

`DraweeView` does extend `ImageView` but as the [documentation on the class says](https://github.com/facebook/fresco/blob/4b94f604e0a98a999dcc1357992f8cb34b1b83b6/drawee/src/main/java/com/facebook/drawee/view/DraweeView.java#L38), that shouldn't be taken as a sign that it supports everything that does.

consp1racy:
DraweeViews support several XML attributes that load images from resources. The best and easiest way to extend functionality would be using

```
fun Context.getDrawableCompat(@DrawableRes resId: Int): Drawable? {
    try {
        // Public API recently introduced in AppCompat.
        return AppCompatResources.getDrawable(this, resId)
    } catch (ex: NoSuchMethodError) {
        try {
            // Private library API so far present in AppCompat at least since 23.4.0.
[...]

PR #2195, opened 2018-08-23
Add support to decode android resources to Drawable directly

zjupure:
## Motivation 

Fresco can only decode resources to Bitmap and not support Drawable. In android platform, drawable is a higher abstraction that can be draw. Sometimes, we want to set a resId in res/drawable to the SimpleDraweeView such as shape drawable or vector drawable, but those cases are not supported by Fresco. Meanwhile, Glide can correctly handle local resources under res/drawable. When Glide found the res can not be decoded into Bitmap, it will try to decode the res to Drawable as a fallback. See https://github.com/bumptech/glide/blob/master/library/src/main/java/com/bumptech/glide/load/resource/drawable/ResourceDrawableDecoder.java#L24  

Though fresco support writing custom decoders to implement the same effect, but it is hard to judge a resId should be decoded to Bitmap or Drawable, because xml file has no major number in header and has been compiled to binary format in apk that can only be recognized by framework. At the same time, when writing a custom decoder to support our own drawable, we have to write at least four classes: a FormatChecker, a CloseableImage, a ImageDecoder, a DrawableFactory.

Some requirement are found in issue #1463 #2173

To solve the existing issue, I have add some code in Fresco to support decode resource to Drawable directly. Changes as following

* Add a `CloseableDrawable` that can wrap the decoded Drawable, this can be returned by our custom decoder. When we write custom decoder, we have no need to implement `CloseableImage` and `DrawableFactory` anymore for basic usage.
* Add a `ResourceDrawableDecoder` to decode a resId to Drawable. The resId can from this app or outer app such as icon.
* Enhance the function of `DefaultImageDecoder` to decode resId to Drawable. When trying to decode an EncodedImage, if we cannot get its image format but we know it is a local resource uri, then we entry a fallback mode and try use `ResourceDrawableDecoder` to decode resId to a Drawable bypass Bitmap. Drawable can also be set into `SimpleDraweeView`.
* Add appcompat-v7 dependency but it is optional by a trick implement. We try to use `AppCompatResources` in support-v7 to getDrawable from resId first because it supports vector drawable for pre-Lollilop devices. If the app do not import support-v7, it will fail once and then use `ContextCompat` in support-v4 to getDrawable. Though Fresco do not want to import support-v7, but there is not very elegant way to work for it.

## Test Plan 

case 1: create a shape drawable xml and load it into SimpleDraweeView.

```
draweeView.setImageUri("res:///"+R.id.shape_dr);
```

case 2: create a vector drawable xml and load it into SimpleDraweeView.

case 3: load an icon from other app into SimpleDraweeView, the icon might be adaptive icon.

```
String pkgName = "com.android.chrome";
PackageManager pm = context.getPackageManager();
ApplicationInfo info = pm.getApplicationInfo(pkgName, 0);
if (info != null && info.icon > 0){
    Uri uri = Uri.parse("android.resources://" + pkgName +"/" + info.icon);
    draweeView.setImageUri(uri);
}
```

case 4: rewrite `SvgDecoder` in `SvgDecoderExample` to return a `ClosableDrawable` and simplify code and it still works well.

```
public static class SvgDecoder implements ImageDecoder {

    @Override
    public CloseableImage decode(
        EncodedImage encodedImage,
        int length,
        QualityInfo qualityInfo,
        ImageDecodeOptions options) {
      try {
        SVG svg = SVG.getFromInputStream(encodedImage.getInputStream());
        return new CloseableDrawable(new SvgPictureDrawable(svg);
      } catch (SVGParseException e) {
        e.printStackTrace();
      }
      return null;
    }
  }
```

Please review the changes and give some suggestions to optimize the issue. I am not concern whether the changes follow the Fresco design or not.

### DIFF
--- a/drawee-backends/drawee-pipeline/src/main/java/com/facebook/drawee/backends/pipeline/DefaultDrawableFactory.java
+++ b/drawee-backends/drawee-pipeline/src/main/java/com/facebook/drawee/backends/pipeline/DefaultDrawableFactory.java
@@ -13,6 +13,7 @@ import android.graphics.drawable.Drawable;
 import android.media.ExifInterface;
 import com.facebook.drawee.drawable.OrientedDrawable;
 import com.facebook.imagepipeline.drawable.DrawableFactory;
+import com.facebook.imagepipeline.image.CloseableDrawable;
 import com.facebook.imagepipeline.image.CloseableImage;
 import com.facebook.imagepipeline.image.CloseableStaticBitmap;
 import com.facebook.imagepipeline.image.EncodedImage;
@@ -54,6 +55,9 @@ public class DefaultDrawableFactory implements DrawableFactory {
               closeableStaticBitmap.getRotationAngle(),
               closeableStaticBitmap.getExifOrientation());
         }
+      } else if (closeableImage instanceof CloseableDrawable) {
+        CloseableDrawable closeableDrawable = (CloseableDrawable)closeableImage;
+        return closeableDrawable.getUnderlyingDrawable();
       } else if (mAnimatedDrawableFactory != null
           && mAnimatedDrawableFactory.supportsImageType(closeableImage)) {
         return mAnimatedDrawableFactory.createDrawable(closeableImage);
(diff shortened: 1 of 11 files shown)


steelrooter:
Stale PR. Closing.

Issue #2171, opened 2018-08-03
Tell Fresco to use disk cache for thumbnails of local videos


hkcorac:
### Description

I'm using version 1.10.0

- In root application, I call `Fresco.initialize(getApplicationContext())`.
- In other UI controls, I pass `Uri.fromFile("file://.../sdcard/example-video.mp4")` to `SimpleDraweeView`. The view generates in-memory thumbnail and displays it correctly.
- I check `/data/data/my.app.id/cache/image_cache/` in emulator, no files or directories are in there. Which means for local videos, Fresco does not use disk cache.

### Additional Information

* Fresco version: 1.10.0
* Platform version: emulator API 23

May I ask if you could consider adding option to tell Fresco to use disk cache for local videos? Normally generating thumbnails takes time, like 1-2 seconds per video (on my real device). So disk cache would help a lot.

Thank you,

lambdapioneer:
Hey @hkcorac, that sounds like a great enhancement. And I'll mark this issue as such.

The thumbnail is created in `LocalVideoThumbnailProducer` and the step where the images are cached on disk is `DiskCacheWriteProducer`. The change would require that `LocalVideoThumbnailProducer` returns an `EncodedImage` instead of an decoded image.

hkcorac:
@lambdapioneer 

Thank you very much. I found `LocalVideoThumbnailProducer` last some days, but couldn't go anywhere more. Your suggestion will help a lot.

PR #2197, opened 2018-08-25
Add option that caches video thumbnails to disk

zjupure:
## Motivation 

As #2171 comments, Fresco loads a thumbnail from local video URI without disk cache. Every time restart app or bitmap in memory cache are evicted, then Fresco need decode thumbnail from video again and it will be slow and poor performance. So put the thumbnail bitmap into disk cache will be a good choice.

To optimize for this case, I create a `LocalVideoThumbnailProducer2` class that will return an `EncodedImage`, and then it can be consumed by `DiskCacheWriteProducer` and `EncodedMemoryCacheProducer`. We can build a new producer sequence for this case, see `ProducerSequenceFactory#getLocalVideoFileFetchSequence2()`.

Of course, it is a feature that you can enable/disable it by `ImageRequestBuilder`, but I think it should be enabled by default.

## Test Plan 

put a video file test.mp4 in sdcard and got the READ_EXTERNAL_STORAGE permission, and use following code to test.

```
File file =  new File(Environment.getExternalStorageDirectory(), "test.mp4");
draweeView.setImageURI(Uri.fromFile(file));
```

### DIFF
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipelineExperiments.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipelineExperiments.java
@@ -21,6 +21,7 @@ import com.facebook.imagepipeline.cache.MemoryCache;
 import com.facebook.imagepipeline.decoder.ImageDecoder;
 import com.facebook.imagepipeline.decoder.ProgressiveJpegConfig;
 import com.facebook.imagepipeline.image.CloseableImage;
+import com.facebook.imagepipeline.memory.BitmapPool;
 import com.facebook.imageutils.BitmapUtil;
 
 /**
@@ -251,6 +252,7 @@ public class ImagePipelineExperiments {
 
     ProducerFactory createProducerFactory(
         Context context,
+        BitmapPool bitmapPool,
         ByteArrayPool byteArrayPool,
         ImageDecoder imageDecoder,
         ProgressiveJpegConfig progressiveJpegConfig,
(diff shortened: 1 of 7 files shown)


erikandre:
Hi @zjupure 

Thank you for your contribution! :)
The idea is good but I think we should be able to do this without creating a separate copy of LocalVideoThumbnailProducer and instead adding this functionality to the existing producer.
What do you think?

zjupure:
@erikandre 

Hi, creating a separate copy of LocalVideoThumbnailProducer is really a bit ugly and the duplicated code might be extracted to an base abstract class.

I think need to create new class not modify the existing class are based on following considration.

* The existing `LocalVideoThumbnailProducer` are return CloseableImage but the new feature requires it return an EncodedImage. 
* If just change the return type of LocalVideoThumbnailProducer, and it will change the whole producer sequence whether the user disable or enable this feature.
* Existing producer sequence:  bitmap cache ---> xxx ---> LocalVideoThumbnailProducer (CloseableImage)
* New feature producer sequence:  bitmap cache ---> DecoderProducer ---> DiskCacheXXXProduer ---> EncodedMemoryCacheProducer ---> LocalVideoThumbnailProducer(EncodedImage)
the new producer sequence work flow: video --> thumnailb bitmap (decoded) ---> inputstream ---> bitmap.  First time without disk cache, it will decoed bitmap twice.  
* If change the existing sequence to this, and user disable the feature, bitmap decoding twice will occur in many cases. In fact, if user disable disk cache, we only need decode once from video file directly.

So I think when user enable this feature and we should build a new producer sequence. But if user disable it using the exsting producer sequence would be a better choice. @erikandre How do you think for this case?

Issue #2201, opened 2018-09-05
Bug when using regiondecode and resize together


qhyuan1992:
when i using regiondecode and resize together, then  i find the image can not show correctly.
`BitmapRegionDecoder` supperts resize.
how to repeat it :
1.init, here `setDownsampleEnabled` is set 
```
ImagePipelineConfig.Builder pipelineConfigureBuilder = ImagePipelineConfig.newBuilder(this);
ImagePipelineConfig pipelineConfig=pipelineConfigureBuilder.setDownsampleEnabled(true).build();
Fresco.initialize(this, pipelineConfig, draweeConfig);
```
2. for a jpg 750*422， ResizeOption(375,211), region(164, 0, 586,422)

what I see is as below，the Bitmap resize did not seem to take effect。Bitmap Size is not 211*211 which I want to be. Also scaleType is not correct.
<img width="150" height="150" src="https://user-images.githubusercontent.com/7428590/45074726-2cdf5c80-b117-11e8-94f6-dd95860f0601.png"/>

whe `setDownsampleEnabled` is set to `true`,  we will use inSampleSize, the following code `targetWidth` and `targetHeight` may be error? 

```
 protected CloseableReference<Bitmap> decodeStaticImageFromStream(
            InputStream inputStream, BitmapFactory.Options options, @Nullable Rect regionToDecode) {
        Preconditions.checkNotNull(inputStream);
        int targetWidth = options.outWidth;
        int targetHeight = options.outHeight;
        if (regionToDecode != null) {
            targetWidth = regionToDecode.width();
            targetHeight = regionToDecode.height();
        }
        // ...
        bitmapToReuse.reconfigure(targetWidth, targetHeight, options.inPreferredConfig);
```

After I change `targetWidth` to  `egionToDecode.width()/options.inSampleSize`, it works correctlly.
<img width="150" height="150" src="https://user-images.githubusercontent.com/7428590/45074752-3ec0ff80-b117-11e8-9923-a4bdb2d61f63.png"/>

lambdapioneer:
@oprisnik , is it a known limitation that resize and region decoding don't work together?

The suggested fix looks correct to me – @qhyuan1992 do you want to submit a PR?

oprisnik:
Yes, I think we didn't properly implement that case yet.  A pull request would be amazing!

qhyuan1992:
ok，I will commit a PR later

qhyuan1992:
[...]

PR #2203, opened 2018-09-06
fix bug when using DownSample and RegionDecoder together

qhyuan1992:
## Motivation

To fix the bug discribed in issue [2201](https://github.com/facebook/fresco/issues/2201)

## Test Plan

The Image can not show correctly when using Resize and RegionDecoder together with `downsample` setting to `true`. Here I change the `targetWidth` and `targetHeight` with region decoder to fix the bug.
when not `setDownsampleEnable` is not  set to `true` or using default value. Resize is a process of pipeline before decoding (only for JPEG). so the region is not the same with the former. but this can be set in custom decoder.


oprisnik:
Thanks for the pull request! We're testing it internally.

PR #2218, opened 2018-09-25
Fix site after enabling SSL

JoelMarcey:

- [ X] Explain the **motivation** for making this change.
- [X ] Provide a **test plan** demonstrating that the code is solid.
- [ X] Match the **code formatting** of the rest of the codebase.
- [ X] Target the `master` branch

## Motivation (required)

Fix website

## Test Plan (required)

Eyes

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,6 +1,7 @@
 ---
 permalink: /blog/:categories/:year/:month/:day/:title.html
-url: "http://frescolib.org"
+url: "https://facebook.github.io/fresco/"
+cdn_url: "https://frescolib.org"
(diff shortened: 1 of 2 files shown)


erikandre:
Thanks for fixing this @JoelMarcey ! 
How long until we can expect the change to reach www.frescolib.org, it still seems broken to me.

JoelMarcey:
@erikandre Hi! Yeah, the site is still broken because shipit is failing due to the CNAME commit that got generated on GitHub only when I removed the custom domain. Once we get shipit fixed (see D10048136 and D10044408) and the commits get pushed, the site should be ok again.

PR #2279, opened 2019-01-24
replace with new domain for Chinese docs

Budaoshi:

- [√] Explain the **motivation** for making this change.
- [√ ] Target the `master` branch

## Motivation (required)

The old Chinese docs site has not been updated for a long time and the docs contents changes more and more. The new one is consistent with the official's.


zmroczek:
Hi @Budaoshi 

Thanks for submitting this and letting us know. We reached out to the person hosting the current Chinese doc web page with a request to update it and would prefer to stay with the old domain if possible. Please let us know if it's still not updated after a while and we will get back to this.

Issue #2318, opened 2019-04-11
Explicit color management


lpy:
### Description

Right now ImagePipeline allows to specify whether to transform to srgb and there are some APIs that use this bit. A more future-proof approach would be explicitly passing a target color space.

Before API level 26, Android only recognized all images as sRGB, right now it looks like old APIs just simply pass a `false` to its new variants to not decode to sRGB, which means the encoded color space of the images would likely be the target color space, applications that use the new version of Fresco and run on new OS might end up not getting what they want if they don't do color management, maybe consider only decode images to sRGB color space for old APIs?

### Solution

Just some thoughts:

1. Consider updating [ImageDecodeOptions](https://github.com/facebook/fresco/blob/master/imagepipeline-base/src/main/java/com/facebook/imagepipeline/common/ImageDecodeOptions.java) to have a ColorSpace, and use it instead of `transformToSRGB`;

2. Update current APIs that accepts `transformToSRGB` to accept a `ColorSpace` instead;

3. Update old APIs (introduced before API level 26) that don't accept `transformToSRGB` to explicit decode images to sRGB.

lpy:
Not sure the backward compatible rules of this project, can API like [DefaultDecoder#decodeJPEGFromEncodedImageWithColorSpace](https://github.com/facebook/fresco/blob/master/imagepipeline/src/main/java/com/facebook/imagepipeline/platform/DefaultDecoder.java#L131) be changed?

zmroczek:
Hi @lpy 

Thanks for suggesting this idea! The approach looks good so far but we would need to discuss more about how to solve it and will come back to you soon.

lpy:
Upload a [pull request](https://github.com/facebook/fresco/pull/2323), please take a review :) If the direction looks good, I can proceed to sign the CLA, thanks!

lpy:
Thanks for taking the pull request. That should cover all places that would use the Android decoder, but I am not sure whether there are other places that we should manage the color. Is there any place that decodes images without using the Android decoder?

lambdapioneer:
@Lpy thanks a lot for doing this! There are two places where we decode without proper colour management, but these places are phasing out slowly:

 - The static-webp backport. This uses libwebp directly on platforms that don't have built-in webp support. This number is pretty low nowadays.
 - The native image resizer uses libjpeg-turbos jpegtran to resize JPEG images. Afaik, it will drop color information. This is also going away mid-term as we want to reduce the amount of native libraries that are coming with Fresco.

PR #2323, opened 2019-04-17
Add explicit color management.

lpy:
Color management was introduced in Android O. Currently Fresco allows
applications to specify whether a decoded bitmap should be transformed to SRGB
color space or not at load time, and legacy APIs pass in false, meaning if
applications don't use newer APIs, the decoded bitmap will have the encoded
color space as the final color space if they have one. This doesn't not align
with the world before Android O because before Android O, everything is
essentially treated as SRGB inside the rendering pipeline no matter whether the
images come with a color space.

This patch, 1) Modifies the newer APIs to accept ColorSpace instead of a
boolean, a more future-proof solution as mobile display technology evolves; 2)
Updates legacy APIs to pass in null, meaning on newer devices (API level >=
26), applications that haven't managed color will always get the decoded bitmap
in SRGB color space.

Issue: #2318

### DIFF
--- a/imagepipeline-base/src/main/java/com/facebook/imagepipeline/common/ImageDecodeOptionsBuilder.java
+++ b/imagepipeline-base/src/main/java/com/facebook/imagepipeline/common/ImageDecodeOptionsBuilder.java
@@ -8,9 +8,11 @@
 package com.facebook.imagepipeline.common;
 
 import android.graphics.Bitmap;
+import android.graphics.ColorSpace;
 import com.facebook.imagepipeline.decoder.ImageDecoder;
 import com.facebook.imagepipeline.transformation.BitmapTransformation;
 import javax.annotation.Nullable;
+import javax.annotation.Nonnull;
 
 /**
  * Builder for {@link ImageDecodeOptions}.
@@ -26,6 +28,7 @@ public class ImageDecodeOptionsBuilder {
   private Bitmap.Config mBitmapConfig = Bitmap.Config.ARGB_8888;
   private @Nullable ImageDecoder mCustomImageDecoder;
   private @Nullable BitmapTransformation mBitmapTransformation;
+  private @Nullable ColorSpace mColorSpace;
 
   public ImageDecodeOptionsBuilder() {
   }
@@ -43,8 +46,8 @@ public class ImageDecodeOptionsBuilder {
     mForceStaticImage = options.forceStaticImage;
     mBitmapConfig = options.bitmapConfig;
     mCustomImageDecoder = options.customImageDecoder;
-    mTransformToSRGB = options.transformToSRGB;
     mBitmapTransformation = options.bitmapTransformation;
+    mColorSpace = options.colorSpace;
     return this;
   }
 
@@ -198,26 +201,6 @@ public class ImageDecodeOptionsBuilder {
   }
 
   /**
-   * Gets whether to allow or not an image color space to be transformed into sRGB.
-   *
-   * @return whether to allow the color space to be transformed into sRGB.
-   */
-  public boolean getTransformToSRGB() {
-    return mTransformToSRGB;
-  }
-
-  /**
-   * Sets whether to allow or not the color space transformation of the image.
-   *
-   * @param transformToSRGB whether to allow the color space to be transformed to sRGB
-   * @return this builder
-   */
-  public ImageDecodeOptionsBuilder setTransformToSRGB(boolean transformToSRGB) {
-    mTransformToSRGB = transformToSRGB;
-    return this;
-  }
-
-  /**
    * Set a custom in-place bitmap transformation that is applied immediately after decoding.
    *
    * @param bitmapTransformation the transformation to use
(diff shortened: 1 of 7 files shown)


imagepipeline-base/src/main/java/com/facebook/imagepipeline/common/ImageDecodeOptionsBuilder.java
```diff
@@ -197,26 +200,6 @@ public ImageDecodeOptionsBuilder setBitmapConfig(Bitmap.Config bitmapConfig) {
...
-    return mTransformToSRGB;
-  }
-
-  /**
-   * Sets whether to allow or not the color space transformation of the image.
-   *
-   * @param transformToSRGB whether to allow the color space to be transformed to sRGB
-   * @return this builder
-   */
-  public ImageDecodeOptionsBuilder setTransformToSRGB(boolean transformToSRGB) {
```
oprisnik: In order to not break the API, we could deprecate this method and just set the color space accordingly (if we decide to support the `null` variant. Otherwise, this will be a no-op I suppose and we can just remove it.

imagepipeline/src/main/java/com/facebook/imagepipeline/nativecode/DalvikPurgeableDecoder.java
```diff
@@ -114,10 +123,14 @@ protected DalvikPurgeableDecoder() {
       Bitmap.Config bitmapConfig,
       @Nullable Rect regionToDecode,
       int length,
-      final boolean transformToSRGB) {
+      @Nullable final ColorSpace colorSpace) {
     BitmapFactory.Options options = getBitmapFactoryOptions(
         encodedImage.getSampleSize(),
         bitmapConfig);
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+      options.inPreferredColorSpace = colorSpace == null
```
oprisnik: This check (and the one above) are going to break DEX preverification. Can you instead move these to a util class, similar to this:

```
    ...
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
      OreoUtils.setColorSpace(options, colorSpace);
    }
    ...
  }



  ....
  import com.facebook.annotations.DoNotOptimize;
  @DoNotOptimize
  private static class OreoUtils {
    void setColorSpace(...) {
        options.inPreferredColorSpace = ...
    }
  }
}
```

imagepipeline/src/main/java/com/facebook/imagepipeline/nativecode/DalvikPurgeableDecoder.java
```diff
@@ -81,10 +84,14 @@ protected DalvikPurgeableDecoder() {
...
       Bitmap.Config bitmapConfig,
       @Nullable Rect regionToDecode,
-      final boolean transformToSRGB) {
+      @Nullable final ColorSpace colorSpace) {
     BitmapFactory.Options options = getBitmapFactoryOptions(
         encodedImage.getSampleSize(),
         bitmapConfig);
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+      options.inPreferredColorSpace = colorSpace == null
+              ? ColorSpace.get(ColorSpace.Named.SRGB) : colorSpace;
```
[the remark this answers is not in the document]
lpy: There's no benefit I can think of to get such behaviour in my humble opinion. One must be very sure about the encoded color space of the contents in order to be confident to skip this bit while also maintain color management in the pipeline. But in that case, the color space is already known in advance, and passing a `ColorSpace` would make the code more readable as well.

imagepipeline-base/src/main/java/com/facebook/imagepipeline/common/ImageDecodeOptionsBuilder.java
lpy: To make sure this library provides benefit of helping manage color, I would suggest removing this, and don't support the `null` variant.

lpy:
Thanks for the feedback, I resolved the last comment and replied the first two comments.

May I ask how to run the tests locally? Android Studio doesn't seem to allow me to run the tests. :) Also, the `showcase` keep returning `error: cannot find symbol class AnimatedDrawable2` in the latest Android Studio :(

oprisnik:
Yeah, the build was broken (I fixed it with 28de385fea88def78bfab4b59d5d7bf10b9ec83e) and our CI is also still acting up. Using Gradle / Android Studio usually works for test.

I'll import your pull request and do some internal testing. Thanks again!

lpy:
Should I wait for your feedback? May I ask help for CI rerun? I am going to squash into one commit to avoid commit pollution as well.

lambdapioneer:
Ipy, don't worry about merging the commits. It will automatically be merged under the PR request including the summary when we land this internally.

When landing we'll also be auto-formatting the change to reflect our code style. This includes flipping some annotations as our tools assume `@Nonnull` be default. After passing our internal tests, this should land today or tomorrow.

Issue #2119, opened 2018-05-25
Local image with exif rotation and ResizeOptions display's wrong


dmitrychistyakov:
### Description

I have image, token with samsung camera and placed on sd-card. File EXIF rotation is ORIENTATION_ROTATE_90 (6). If I add ResizeOptions to ImageRequestBuilder, the image looks compressed by vertical.

### Reproduction

Here is the sample file
https://www.dropbox.com/s/fsixt3lfbw6zqiv/Exif270.jpg?dl=0
Place it to sd-card root

And here is code

```
public class MainActivity extends AppCompatActivity {

    private static String[] PERMISSIONS_STORAGE = {
            Manifest.permission.READ_EXTERNAL_STORAGE,
            Manifest.permission.WRITE_EXTERNAL_STORAGE
    };

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);

        Fresco.initialize(this);

        verifyStoragePermissions(this);

        SimpleDraweeView simpleDraweeView = new SimpleDraweeView(this);
        simpleDraweeView.setLayoutParams(new ViewGroup.LayoutParams(
                ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.MATCH_PARENT));
        File file = new File("/storage/emulated/0/Exif270.jpg");
        ImageRequestBuilder imageRequestBuilder = ImageRequestBuilder.newBuilderWithSource(
                Uri.fromFile(file)).setResizeOptions(ResizeOptions.forDimensions(100, 300));
        PipelineDraweeControllerBuilder controllerBuilder = Fresco.newDraweeControllerBuilder()
                .setImageRequest(imageRequestBuilder.build());
        simpleDraweeView.setController(controllerBuilder.build());
        setContentView(simpleDraweeView);
    }

public static void verifyStoragePermissions(Activity activity) {
        int permission = ActivityCompat.checkSelfPermission(activity, Manifest.permission.WRITE_EXTERNAL_STORAGE);
        if (permission != PackageManager.PERMISSION_GRANTED) {
            ActivityCompat.requestPermissions(activity, PERMISSIONS_STORAGE, 1);
        }
    }
}

```

Screenshot of result
![resize-bug](https://user-images.githubusercontent.com/2329931/40545518-cc1189f6-6034-11e8-8575-689f789a2c7a.png)

Without resizing everything works fine

### Additional Information

* Fresco version: 1.9.0
[...]

PR #2383, opened 2019-07-25
Fix thumbnail image orientation

dimach1977:
Summary:
Some images are displayed with an incorrect orientation. The orientation in Exif is correct and the image itself is just a selfie. Another selfie from the same phone is shown correctly and generally about 2 selfies out of 10 are shown with the wrong orientation.

I'm not completely certain how this happens, but there seems like a race condition between caching and resizing: in some images EncodedImage.parseMetaData gets to work on the full image, so it is able to read the orientation value from Exif, while for some images, parseMetaData is called when the image already has mRotationAngle set (by going through copyMetaDataFrom from the just-made cache maybe?)

Regardless, it seems logical that the image orientation should only be reset if it isn't set already.

This possibly fixes this issue as well: https://github.com/facebook/fresco/issues/2119

Reviewed By: oprisnik

Differential Revision: D16460898


oprisnik:
This was landed already.

PR #2385, opened 2019-07-30
Change method name 'generate' to 'createJpeg', 'increase' to 'includeBitmap', 'ensure' to 'loadLibrary', 'build' to 'getImageRequest', 'get' to 'createContext'

pdhung3012:
The class EmptyJpegGenerator is used to generate an empty JPEG image.  This method named "generate" is to create a new JPEG image. Thus, the method name "createJpeg" is more intuitive than "generate".

The class BitmapCounter is used to count the Bitmaps.  This method named "increase" is to check the bitmap. Thus, the method name "includeBitmap" is more intuitive than "increase".

The class BitmapCounter is used to represent GifImage.  This method named 'ensure' is to load library for image processing. Thus, the method name 'loadLibrary' is more intuitive than 'ensure'.

The class ImageUrlsRequestBuilder is used to build ImageUrlsRequest.  This method named 'build' is to get a new Image url request. Thus, the method name 'getImageRequest' is more intuitive than 'build'.

The class DefaultFrescoContext  is used to represent DefaultFrescoContext.  This method named 'get' is to create a context. Thus, the method name 'createContext' is more precise than 'get'.

### DIFF
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/memory/BitmapCounter.java
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/memory/BitmapCounter.java
@@ -8,16 +8,11 @@
 package com.facebook.imagepipeline.memory;
 
 import android.graphics.Bitmap;
-import android.os.Build;
+
 import com.facebook.common.internal.Preconditions;
-import com.facebook.common.internal.Throwables;
-import com.facebook.common.references.CloseableReference;
 import com.facebook.common.references.ResourceReleaser;
-import com.facebook.imagepipeline.common.TooManyBitmapsException;
-import com.facebook.imagepipeline.nativecode.Bitmaps;
 import com.facebook.imageutils.BitmapUtil;
-import java.util.ArrayList;
-import java.util.List;
+
 import javax.annotation.concurrent.GuardedBy;
 
 /**
@@ -59,7 +54,7 @@ public class BitmapCounter {
    * @param bitmap to include in the count
    * @return true if and only if bitmap is successfully included in the count
    */
-  public synchronized boolean increase(Bitmap bitmap) {
+  public synchronized boolean isInBitmap(Bitmap bitmap) {
     final int bitmapSize = BitmapUtil.getSizeInBytes(bitmap);
     if (mCount >= mMaxCount || mSize + bitmapSize > mMaxSize) {
       return false;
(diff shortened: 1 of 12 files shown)


vito/provider/src/main/java/com/facebook/fresco/vito/provider/DefaultFrescoContext.java
```diff
@@ -28,7 +28,7 @@
   private static @Nullable FrescoContext sInstance;
   private static @Nullable Supplier<Boolean> sDebugOverlayEnabledSupplier;
 
-  public static synchronized FrescoContext get(Resources resources) {
+  public static synchronized FrescoContext createContext(Resources resources) {
```
oprisnik: This change doesn't make sense. It doesn't always create a new instance.

samples/comparison/src/main/java/com/facebook/samples/comparison/urlsfetcher/ImageUrlsRequestBuilder.java
```diff
@@ -39,7 +39,7 @@ public ImageUrlsRequestBuilder addImageFormat(ImageFormat imageFormat, ImageSize
     return this;
   }
 
-  public ImageUrlsRequest build() {
+  public ImageUrlsRequest getImageRequest() {
```
oprisnik: This doesn't make sense and is a huge breaking API change.

imagepipeline/src/main/java/com/facebook/imagepipeline/memory/BitmapCounter.java
```diff
@@ -59,7 +54,7 @@ public void release(Bitmap value) {
    * @param bitmap to include in the count
    * @return true if and only if bitmap is successfully included in the count
    */
-  public synchronized boolean increase(Bitmap bitmap) {
+  public synchronized boolean includeBitmap(Bitmap bitmap) {
```
oprisnik: The old name is clearer IMO.

pdhung3012: Thank you for your notification. Do you think that, the other case 'generate' to 'createJpeg' improve the readability?

defHLT:
It's unclear that the proposed names are better. Closing as it has been stale for a long time.

Issue #2426, opened 2019-10-28
Showcase crash for API <21


Yithis:
### Description

Showcase is crashing for API <21.

### Reproduction
Open "Placeholder, Progress, Failure" tab.
https://imgur.com/a/eyfzeOv

### Solution

https://stackoverflow.com/a/41629944
https://imgur.com/a/yWTwNrl

### Additional Information
* Platform version: [API <21]

oprisnik:
Thanks! The PR has been merged.

PR #2427, opened 2019-10-28
Fix crash for API <21

Yithis:
## Motivation (required)

https://github.com/facebook/fresco/issues/2426

## Test Plan (required)
Running "Showcase" app.
`./gradlew assembleDebug`


oprisnik:
Awesome, thanks @Yithis!

Issue #2397, opened 2019-08-22
Low image quality using <Image/> component on RN > = 0.57 (Fresco >= 1.10.0)


clytras:
### Description

RN issue: [RN 0.57.x Bundled large images have low quality when viewing using \<Image/\> component with 1:1 AR on Android [CLOSED]](https://github.com/facebook/react-native/issues/21301)

There is low quality when loading large bundled (PNG, GIF and maybe more formats, **NOT JPEG**) images only on Android:

<img src="https://zikro.gr/img/RN_0_56_0_BigImageSource.png" width="310"/> <img src="https://zikro.gr/img/RN_0_57_1_BigImageSource.png" width="310"/>

At the left screenshot we see the exact same code running with **RN 0.56.0** and at the right screenshot we see **RN 0.57.1**. The code is just a simple image `<Image source={require('./assets/ELHall1.png')} resizeMethod="resize" />` and the image size is `2111 x 4645 pixels`. Both projects are fresh installed using `react-native init RN057ImageTest` and `react-native init --version="0.56.0" RN056ImageTest`. This continues to happen from 0.56 and all versions after and latest RN 0.60.x.

This is confirmed to be caused by RN Fresco lib change between 0.56 and 0.57 from 1.9.0 to 1.10.0 https://github.com/facebook/react-native/commit/b6f2aad9c0119d11e52978ff3fa9c6f269f04a14. Check comment https://github.com/facebook/react-native/issues/21301#issuecomment-520155609.

After some search at Fresco issues, I can see some related issues that it's suggested that large images should be divided and recomposed piece by piece, which it resolves many cases (most map related large images) but it can be very inconvenient especially for dynamic loaded/created images. This was working until RN 0.56 and from 0.57 and after it does not.

### Reproduction

RN: It's the initial `App.js` with an `<Image/>` component added.
```
...
type Props = {};
export default class App extends Component<Props> {
  render() {
    return (
      <View style={styles.container}>
        <Image source={require('./assets/ELHall1.png')} resizeMethod="resize" />
      </View>
    );
  }
}
...
```

### Additional Information

At this comment https://github.com/facebook/react-native/issues/21301#issuecomment-520418832, [lambdapioneer](https://github.com/lambdapioneer) writes that this maybe is related to scale down (sub-sampling) large images:

[...]

PR #2500, opened 2020-05-10
Respect setDownsampleEnabled config in DecodeProducer

sunnylqm:
Fixes https://github.com/facebook/fresco/issues/2397

## Motivation (required)

Currently `setDownsampleEnabled` is not properly applied if set to false.

## Test Plan (required)

My understanding is that `setDownsampleEnabled(false)` __should disable__ this feature no matter what other conditions are, while `setDownsampleEnabled(true)` __may enable it__ or leave it to other conditions.


JeffGuKang:
It there any update about this?

sunnylqm:
ping @defHLT  @oprisnik

pateljoel:
ping @defHLT @oprisnik

pateljoel:
@passy @oprisnik  @defHLT not sure if pinging is working but if anyone at Facebook can look into this it would be greatly appreciated as this is a serious issue in fresco.

pateljoel:
@passy @oprisnik @defHLT gentle ping

sunnylqm:
Is there a change we get some love & attention from RN team to this little 4-year-old issue?   @cortinico @javache @mdvacca @yungsters 

some context here: 
https://github.com/facebook/fresco/issues/2397#issuecomment-543223113

pateljoel:
gentle ping @cortinico @javache @mdvacca @yungsters @passy @oprisnik @defHLT

cortinico:
> if there are confidential concerns, i don't know what else to do except update the pr😂

FYI: Updating the PR is not helpful. @defHLT should drive this forward as I mentioned already

pateljoel:
Apologies for asking, but has there been any progress internally about merging in this issue yet? 

If the team at Facebook need a reproducible example of this issue please look at the issue below.

https://github.com/facebook/fresco/pull/2500#issuecomment-1698751392

yungsters:
The primary concern with this change is the likelihood of unintended downstream impact it may have on other use cases — not employing React Native. However, this is supposedly how Fresco has always worked since 2017: https://github.com/facebook/fresco/commit/8d83ed272a0b9602214df8b1004430de302d34cc

A couple potential ideas for next steps:

- Figure out why downsampling produces such a poor result. It should not result in an image with lower dimensions than requested. @defHLT suggested that this method might warrant some investigation: https://github.com/facebook/fresco/blob/a4ac78a413bb8e5eef6339df0a9a11aadcaed64f/imagepipeline-base/src/main/java/com/facebook/imagepipeline/transcoder/DownsampleUtil.kt#L118-L131
- Or determine why this started regressing in React Native and identify a mitigation based on that.

sunnylqm:
Even if it `may` impact other use cases, it is quite common to bump a major version with some notes about the breaking change, instead of leaving it rotting and `substantially impacting` react native users for years. 
Do we have an option to disable downsampling regardless how smart it is? The `setDownsampleEnabled` simply not works, like a door you can only open but not close —— it's a bug that should be fixed in every sense.
And I don't think the downsampling performance matters here. We all know it's a tradeoff. The key point of a tradeoff is not about the price/performance, it's all about choice. Even if you can achieve 99.9% performance, I still have the right to choose a lossless way right?

yungsters:
@sunnylqm – I understand your frustration, and I generally agree with what you're saying. (Also, I really appreciate the folks who have been pinging this and ensuring it gets the attention that it deserves.)

When I previously mentioned the following:

> The primary concern with this change is the likelihood of unintended downstream impact it may have on other use cases — not employing React Native.

I was referring to other use cases internally at Meta. Like React Native, Fresco is also used in production apps (and at the scale that these apps operate at, almost anything that can go wrong often does to some degree).

Rest assured, I will continue to follow up on this internally.

yungsters:
I just wanted to follow up here so everyone knows that this is still on my mind.

I've proposed a couple alternatives to the team, and I will report back with any updates.

yungsters:
Thanks again for the patience. I've shared a proposal with other engineers at Meta that everyone is happy to support, and I believe it will fulfill the requirements here.

Instead of changing the existing behavior of `setDownsampleEnabled(false)`, which is too risky of a change that is depended upon by many existing apps, I propose adding a new method and deprecating `setDownsampleEnabled(boolean)` entirely. The premise is that `setDownsampleEnabled(boolean)` is the wrong interface anyway, because `setDownsampleEnabled(false)` is misleading and does not actually do what most people would expect.

We can introduce a new method, `setDownsampleMode(DownsampleMode)`, where `DownsampleMode` is an enum with values:

  * `DownsampleMode.ALWAYS` behaves the same as today's `setDownsampleEnabled(true)`.
  * `DownsampleMode.AUTO` behaves the same as today's `setDownsampleEnabled(false)`.
  * `DownsampleMode.NEVER` is new and will cause Fresco to ignore the `IS_RESIZING_DONE` flag.
  
We can then deprecate `setDownsampleEnabled(boolean)` and refactor it to invoke `setDownsampleMode` for backward compatibility. The check in `DecodeProducer.java` that is being changed in this PR can then become:

```
if (mDownsampleMode != DownsampleMode.NEVER || !statusHasFlag(status, Consumer.IS_RESIZING_DONE)) {
```

Does this make sense and solve the problems that this PR is trying to solve? And if so, would this proposal be something that you would be willing to help implement?

sunnylqm:
thank you for your proposal and it sounds perfect. will try to implement in a few days

pateljoel:
@yungsters Sounds good, so if this is implemented by @sunnylqm does this mean it would be merged in quicker? Would this also get resolved for React Native as well as fresco is depended on for image loading?

yungsters:
Yes, reviewing the new proposal will be faster. And yes, any consumers (including React Native) will benefit from the changes.

sunnylqm:
@yungsters please review https://github.com/facebook/fresco/pull/2760

PR #2501, opened 2020-05-15
fix #2497

### LINKED ISSUE #2497: Callbacks do not work when both `setLowResImageRequest` and `setImageRequest` are used
### Description

```
val normalUriString = "https://www.google.com.hk/logos/doodles/2020/stay-and-play-at-home-with-popular-past-google-doodles-halloween-2016-6753651837108773-s.png"
val 404UriString = "https://www.google.com.hk/404"

vImage.controller = Fresco.newDraweeControllerBuilder()
     .setLowResImageRequest(ImageRequest.fromUri(Uri.parse(normalUriString)))
     .setImageRequest(ImageRequest.fromUri(Uri.parse(404UriString)))
     .setControllerListener(object : BaseControllerListener<ImageInfo>() {
         override fun onFailure(id: String?, throwable: Throwable?) {
             Log.e("TAG", "onFailure")
         }

         override fun onIntermediateImageSet(id: String?, imageInfo: ImageInfo?) {
             Log.e("TAG", "onIntermediateImageSet")
         }

         override fun onFinalImageSet(id: String?, imageInfo: ImageInfo?, animatable: Animatable?) {
             Log.e("TAG", "onFinalImageSet")
         }
     })
     .setOldController(vImage.controller)
     .build()

fun clearCache() {
     Fresco.getImagePipeline().clearCaches()
}
```

expect：
```
E/TAG: onIntermediateImageSet
E/TAG: onFailure
```

actually：
It works the first time, but after calling `clearCache` **a few times**(call several times), It doesn't work：

```
E/TAG: onFinalImageSet
```

### Reproduction
As shown above

jokermonn:
## Motivation

fix #2497 

image

The reason for the problem is that runnable's execution is asynchronous, and `com.facebook.datasource.AbstractDataSource#mDataSourceStatus` may change during this time.

Because the first execution to [this line](https://github.com/facebook/fresco/blob/master/fbcore/src/main/java/com/facebook/datasource/AbstractDataSource.java#L193) will eventually result in the mDataSource being [null](https://github.com/facebook/fresco/blob/master/drawee/src/main/java/com/facebook/drawee/controller/AbstractDraweeController.java#L628), even though the two fields are the same on the second execution (line 5 of the log), the method returns false because the mDataSource is null, resulting in the onFailure method not being called([this method](https://github.com/facebook/fresco/blob/master/drawee/src/main/java/com/facebook/drawee/controller/AbstractDraweeController.java#L602) will return false).

### DIFF
--- a/fbcore/src/main/java/com/facebook/datasource/AbstractDataSource.java
+++ b/fbcore/src/main/java/com/facebook/datasource/AbstractDataSource.java
@@ -164,30 +164,26 @@ public abstract class AbstractDataSource<T> implements DataSource<T> {
     }
 
     if (shouldNotify) {
-      notifyDataSubscriber(dataSubscriber, executor, hasFailed(), wasCancelled());
+      notifyDataSubscriber(dataSubscriber, executor);
     }
   }
 
   private void notifyDataSubscribers() {
-    final boolean isFailure = hasFailed();
-    final boolean isCancellation = wasCancelled();
     for (Pair<DataSubscriber<T>, Executor> pair : mSubscribers) {
-      notifyDataSubscriber(pair.first, pair.second, isFailure, isCancellation);
+      notifyDataSubscriber(pair.first, pair.second);
     }
   }
 
   protected void notifyDataSubscriber(
       final DataSubscriber<T> dataSubscriber,
-      final Executor executor,
-      final boolean isFailure,
-      final boolean isCancellation) {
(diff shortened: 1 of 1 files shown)


jokermonn:
？

jokermonn:
@oprisnik

jokermonn:
so, does it can not be merged?

PR #2514, opened 2020-07-22
permit developer customizing a implementation of  SerialExecutorService

NickHu150:
## Motivation (required)
We found that fresco has a serious performance problem while side RecyclerView with a large number of animed webps.Then we check the source code and found the maxConcurrency in class DefaultSerialExecutorService is too small to excute tasks relate to (frame.renderFrame)

## Test Plan (required)
use foloowing url and you can recur issue http://cdn.poizon.com/algo/webp/62e595747d7279bbb497026487e1f735_dur124dur_1589537743359_w1080h1920_w480h672.webp 

new initialized code which can slove the issue

```
    val supplier = DefaultExecutorSupplier2(5)
    val pipelineConfig = ImagePipelineConfig.newBuilder(this)
            .setExecutorSupplier(supplier)
            .setSerialExecutorService(DefaultSerialExecutorService2(supplier.forDecode()))
        .build()
```

```
public class DefaultExecutorSupplier2 implements ExecutorSupplier {
  // Allows for simultaneous reads and writes.
  private static final int NUM_IO_BOUND_THREADS = 20;
  private static final int NUM_LIGHTWEIGHT_BACKGROUND_THREADS = 10;
……
```

```
public class DefaultSerialExecutorService2 extends ConstrainedExecutorService
    implements SerialExecutorService {

  public DefaultSerialExecutorService2(Executor executor) {
     // change the maxConcurrency to 5 or higher can improve the sliding performance clearly
    super("SerialExecutor", 5, executor, new LinkedBlockingQueue<Runnable>());
  }
```

### DIFF
--- a/animated-base/src/main/java/com/facebook/fresco/animation/factory/AnimatedFactoryV2Impl.java
+++ b/animated-base/src/main/java/com/facebook/fresco/animation/factory/AnimatedFactoryV2Impl.java
@@ -54,7 +54,7 @@ public class AnimatedFactoryV2Impl implements AnimatedFactory {
   private @Nullable AnimatedDrawableBackendProvider mAnimatedDrawableBackendProvider;
   private @Nullable AnimatedDrawableUtil mAnimatedDrawableUtil;
   private @Nullable DrawableFactory mAnimatedDrawableFactory;
-  SerialExecutorService mSerialExecutorService;
+  private @Nullable SerialExecutorService mSerialExecutorService;
 
   @DoNotStrip
   public AnimatedFactoryV2Impl(
@@ -67,7 +67,7 @@ public class AnimatedFactoryV2Impl implements AnimatedFactory {
     mExecutorSupplier = executorSupplier;
     mBackingCache = backingCache;
     mDownscaleFrameToDrawableDimensions = downscaleFrameToDrawableDimensions;
-    this.mSerialExecutorService = serialExecutorServiceForFramePreparing;
+    mSerialExecutorService = serialExecutorServiceForFramePreparing;
   }
 
   @Nullable
--- a/animated-base/src/main/java/com/facebook/fresco/animation/factory/AnimatedFactoryV2Impl.java
+++ b/animated-base/src/main/java/com/facebook/fresco/animation/factory/AnimatedFactoryV2Impl.java
@@ -54,17 +54,20 @@ public class AnimatedFactoryV2Impl implements AnimatedFactory {
   private @Nullable AnimatedDrawableBackendProvider mAnimatedDrawableBackendProvider;
   private @Nullable AnimatedDrawableUtil mAnimatedDrawableUtil;
   private @Nullable DrawableFactory mAnimatedDrawableFactory;
+  SerialExecutorService mSerialExecutorService;
 
   @DoNotStrip
   public AnimatedFactoryV2Impl(
       PlatformBitmapFactory platformBitmapFactory,
       ExecutorSupplier executorSupplier,
       CountingMemoryCache<CacheKey, CloseableImage> backingCache,
-      boolean downscaleFrameToDrawableDimensions) {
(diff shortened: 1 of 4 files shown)


animated-base/src/main/java/com/facebook/fresco/animation/factory/AnimatedFactoryV2Impl.java
```diff
@@ -54,17 +54,20 @@
   private @Nullable AnimatedDrawableBackendProvider mAnimatedDrawableBackendProvider;
   private @Nullable AnimatedDrawableUtil mAnimatedDrawableUtil;
   private @Nullable DrawableFactory mAnimatedDrawableFactory;
+  SerialExecutorService mSerialExecutorService;
```
oprisnik, line 57: can you please add `private @Nullable`

animated-base/src/main/java/com/facebook/fresco/animation/factory/AnimatedFactoryV2Impl.java
```diff
@@ -54,17 +54,20 @@
...
       ExecutorSupplier executorSupplier,
       CountingMemoryCache<CacheKey, CloseableImage> backingCache,
-      boolean downscaleFrameToDrawableDimensions) {
+      boolean downscaleFrameToDrawableDimensions,
+      SerialExecutorService serialExecutorServiceForFramePreparing) {
     mPlatformBitmapFactory = platformBitmapFactory;
     mExecutorSupplier = executorSupplier;
     mBackingCache = backingCache;
     mDownscaleFrameToDrawableDimensions = downscaleFrameToDrawableDimensions;
+    this.mSerialExecutorService = serialExecutorServiceForFramePreparing;
```
oprisnik, line 70: `this.` is not required.

imagepipeline/src/main/java/com/facebook/imagepipeline/core/ImagePipelineConfig.java
```diff
@@ -699,6 +707,11 @@ public Builder setEncodedMemoryCache(
       return this;
     }
 
+    public Builder setSerialExecutorService(@Nullable SerialExecutorService serialExecutorService){
```
oprisnik, line 710: The name here is very generic, so it's hard to see what this is used for. Can you name this something like setExecutorServiceForAnimatedImages or similar (maybe even include `preparing`).

animated-base/src/main/java/com/facebook/fresco/animation/factory/AnimatedFactoryV2Impl.java
oprisnik: Also, I don't think this has to be a `SerialExecutorService` but can be anything. It's just that the default implementation is one. We can just pass down an `ExecutorService` to make this even more flexible.

imagepipeline-base/src/main/java/com/facebook/imagepipeline/animated/factory/AnimatedFactoryProvider.java
```diff
@@ -24,7 +25,8 @@ public static AnimatedFactory getAnimatedFactory(
       PlatformBitmapFactory platformBitmapFactory,
       ExecutorSupplier executorSupplier,
       CountingMemoryCache<CacheKey, CloseableImage> backingCache,
-      boolean downscaleFrameToDrawableDimensions) {
+      boolean downscaleFrameToDrawableDimensions,
+      SerialExecutorService serialExecutorService) {
```
oprisnik, line 29: This can then also be just an ExecutorService

NickHu150:
the issues have been fixed

NickHu150:
@oprisnik  please ~

oprisnik (approved):
Thank you, I'll import this for internal testing.

PR #2518, opened 2020-07-27
Updated CONTRIBUTING.md

chaytanyasinha:
Made it much more easy to be understood for new users.

I believed that it will be much easier for the contributors who are new to Fresco.


oprisnik:
Thank you! Looks good

oprisnik:
After discussing this with the team, we don't see the advantage of changing the wording here.

Issue #2499, opened 2020-05-06
fbcore's proguard limit is too broad


NormanOu:
### Description

the proguard configuration in 'fbcore-2.1.0.aar' includes a line that keeps all classes with a native method, which is not good. As we can't choose to shrink other classes with native method any more.

` # Keep native methods
 -keepclassmembers class * {
     native <methods>;
 }`

### Reproduction

any demo app implements fresco should be able to reproduce

### Solution

change the proguard rule to be something like the following
` # Keep native methods
 -keepclassmembers class com.facebook.** {
     native <methods>;
 }`

### Additional Information

* Fresco version: 2.1.0
* Platform version: android

oprisnik:
Thanks, this should be fixed. Pull requests are welcome! :)

PR #2519, opened 2020-08-10
narrow fbcore proguard rule

esafirm:
## Motivation (required)

Close https://github.com/facebook/fresco/issues/2499

## Test Plan (required)

1. Create `aar` and do `unzip -l fbcore.aar` to check if the rule is already changed
2. Build `normalRelease` flavour of `samples:showcase` and test some features
3. Build `samples:animation2` build type `debug` and minify enabled

For 2 and 3 I added `proguard-rules.pro` in `proguardFiles` just to make sure the proguard from aar is included

### DIFF
--- a/fbcore/proguard-fresco.pro
+++ b/fbcore/proguard-fresco.pro
@@ -16,7 +16,7 @@
 }
 
 # Keep native methods
--keepclassmembers class * {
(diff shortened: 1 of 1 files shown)


oprisnik:
Thanks, looks good :)

oprisnik (approved):

Issue #2460, opened 2020-02-07
Unreleased Resource: Streams


QiAnXinCodeSafe:
https://github.com/facebook/fresco/blob/c00077a4311fa220548a7ec66038c146f04d3ceb/imagepipeline/src/main/java/com/facebook/imagepipeline/platform/DefaultDecoder.java#L142

The program can potentially fail to release a system resource.

PR #2520, opened 2020-08-10
close jpegDataStream in finally

esafirm:
## Motivation (required)

Fix https://github.com/facebook/fresco/issues/2460

## Test Plan (required)

Test out the show case app on random uri menu (since it loads JPEG images)


wizh:
Hi, thanks for the fix!

Issue #2504, opened 2020-05-26
Fatal Exception: java.lang.NullPointerException: Attempt to get length of null array


pnkshir:
We are getting following crash on Crashlytics for good number of users with Android OS 8, with more than 47 devices.
We are using "com.facebook.fresco:fresco:2.2.0"

Could you please help us, since its happening on. multiple users.

Logs : 

Fatal Exception: java.lang.NullPointerException: Attempt to get length of null array
       at java.util.Formatter$FormatSpecifier.addZeros(Formatter.java:3505)
       at java.util.Formatter$FormatSpecifier.print(Formatter.java:3401)
       at java.util.Formatter$FormatSpecifier.print(Formatter.java:3332)
       at java.util.Formatter$FormatSpecifier.print(Formatter.java:3317)
       at java.util.Formatter$FormatSpecifier.printFloat(Formatter.java:2891)
       at java.util.Formatter$FormatSpecifier.print(Formatter.java:2844)
       at java.util.Formatter.format(Formatter.java:2523)
       at java.util.Formatter.format(Formatter.java:2458)
       at java.lang.String.format(String.java:2814)
       at com.facebook.common.logging.FLog.formatString(FLog.java:491)
       at com.facebook.common.logging.FLog.v(FLog.java:118)
       at com.facebook.imagepipeline.transcoder.DownsampleUtil.determineDownsampleRatio(DownsampleUtil.java:87)
       at com.facebook.imagepipeline.transcoder.DownsampleUtil.determineSampleSize(DownsampleUtil.java:43)
       at com.facebook.imagepipeline.producers.DecodeProducer$ProgressiveDecoder$1.run(DecodeProducer.java:163)
       at com.facebook.imagepipeline.producers.JobScheduler.doJob(JobScheduler.java:202)
       at com.facebook.imagepipeline.producers.JobScheduler.access$000(JobScheduler.java:22)
       at com.facebook.imagepipeline.producers.JobScheduler$1.run(JobScheduler.java:73)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
       at com.facebook.imagepipeline.core.PriorityThreadFactory$1.run(PriorityThreadFactory.java:51)
       at java.lang.Thread.run(Thread.java:764)

oprisnik:
Thanks for the bug report.

[...]

PR #2552, opened 2020-10-24
Add DecimalFormat for formatting of the float values to string

tsybanov:
## Motivation (required)

Closes (https://github.com/facebook/fresco/issues/2504)
Details on the fix (https://github.com/facebook/fresco/issues/2504#issuecomment-657771489)

This PR fixes the local problem only. I believe it would be better to preformat float values in strings within `Flog.formatString()` by using `DecimalFormat`. However, in that case, it would increase complexity since it'd require placeholders parsing.

## Test Plan (required)

Since the logic hasn't changed, I believe using existing tests is enough. Due to the fact that the problem occurs on an unknown list of OS, I wasn't able to find an edge case to create a test.

### DIFF
--- a/imagepipeline-base/src/main/java/com/facebook/imagepipeline/transcoder/DownsampleUtil.java
+++ b/imagepipeline-base/src/main/java/com/facebook/imagepipeline/transcoder/DownsampleUtil.java
@@ -15,6 +15,7 @@ import com.facebook.imagepipeline.common.ResizeOptions;
 import com.facebook.imagepipeline.common.RotationOptions;
 import com.facebook.imagepipeline.image.EncodedImage;
 import com.facebook.infer.annotation.Nullsafe;
+import java.text.DecimalFormat;
 import javax.annotation.Nullable;
 
 @Nullsafe(Nullsafe.Mode.STRICT)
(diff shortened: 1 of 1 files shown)


imagepipeline-base/src/main/java/com/facebook/imagepipeline/transcoder/DownsampleUtil.java
```diff
@@ -15,6 +15,7 @@
 import com.facebook.imagepipeline.common.RotationOptions;
 import com.facebook.imagepipeline.image.EncodedImage;
 import com.facebook.infer.annotation.Nullsafe;
+import java.text.DecimalFormat;
```
daddivan69, line 18:
```suggestion
import java.text.DecimalFormat;
```suggestion
import java.text.DecimalFormat;

```

```

**Review by daddivan69 — APPROVED** [review]

File: imagepipeline-base/src/main/java/com/facebook/imagepipeline/transcoder/DownsampleUtil.java [hunk 2]
```diff
@@ -15,6 +15,7 @@
 import com.facebook.imagepipeline.common.RotationOptions;
 import com.facebook.imagepipeline.image.EncodedImage;
```
daddivan69, line 16:
```suggestion
 import com.facebook.imagepipeline.image.EncodedImage;
```

imagepipeline-base/src/main/java/com/facebook/imagepipeline/transcoder/DownsampleUtil.java
```diff
@@ -15,6 +15,7 @@
...
 import com.facebook.infer.annotation.Nullsafe;
+import java.text.DecimalFormat;
 import javax.annotation.Nullable;
 
 @Nullsafe(Nullsafe.Mode.STRICT)
```
daddivan69, line 21:
```suggestion
 @Nullsafe(Nullsafe.Mode.)
```

daddivan69 (approved):

imagepipeline-base/src/main/java/com/facebook/imagepipeline/transcoder/DownsampleUtil.java
```diff
@@ -102,17 +103,22 @@ public static float determineDownsampleRatio(
     final float widthRatio = ((float) resizeOptions.width) / widthAfterRotation;
     final float heightRatio = ((float) resizeOptions.height) / heightAfterRotation;
     float ratio = Math.max(widthRatio, heightRatio);
+
+    // Format float values for the log message in advance
+    // due to possible String.Format() NPE within FLog.formatString() in some OS
+    final DecimalFormat dimensionsFormatter = new DecimalFormat("#.#");
```
daddivan69, line 109:
```suggestion
    final DecimalFormat dimensionsFormatter = new DecimalFormat("#.#");
```

PR #2579, opened 2021-01-20
Add synchronized to help prevent race condition with fresco client code

alsutton:

- [X] Explain the **motivation** for making this change.
- [X] Provide a **test plan** demonstrating that the code is solid.
- [X] Match the **code formatting** of the rest of the codebase.
- [X] Target the `master` branch

## Motivation (required)

There is a period of time between checking if NativeLoader has been initialized, and when we try to perform `.init()` which means a race condition can occur (as mentioned in https://github.com/facebook/SoLoader/issues/60). This PR provides a mechanism for client code to avoid that race condition.

This change uses synchronization on the `NativeLoader` class object, and a secondary check to verify that the initialization needs to take place just before we perform it in order to reduce the window in which a race condition can happen.

When combined with https://github.com/facebook/SoLoader/pull/69 this will remove the race condition from the code because the initialisation inside SoLoader and inside Fresco will both be synchronized on the `NativeLoader.class` object. Without that PR in SoLoader this change will still allow fresco clients to synchronize on `NativeLoader.class` in their own code around any areas which may race with this code in fresco.

## Test Plan (required)

This code doesn't alter the functionality of fresco, and so existing tests should provide verification that there is no operational impact.

## Next Steps

Sign the [CLA][2], if you haven't already.

Small pull requests are much easier to review and more likely to get merged. Make sure the PR does only one thing, otherwise please split it.

Make sure all **tests pass** on [Circle CI][4]. PRs that break tests are unlikely to be merged.

For more info, see the [Contributing guide][4].

[1]: https://medium.com/@martinkonicek/what-is-a-test-plan-8bfc840ec171#.y9lcuqqi9
[2]: https://code.facebook.com/cla
[3]: http://circleci.com/gh/facebook/fresco
[4]: https://github.com/facebook/fresco/blob/master/CONTRIBUTING.md

### DIFF
--- a/drawee-backends/drawee-pipeline/src/main/java/com/facebook/drawee/backends/pipeline/Fresco.java
+++ b/drawee-backends/drawee-pipeline/src/main/java/com/facebook/drawee/backends/pipeline/Fresco.java
@@ -84,16 +84,16 @@ public class Fresco {
         clazz.getMethod("init", Context.class).invoke(null, context);
       } catch (ClassNotFoundException e) {
         // Failed to initialize SoLoader
-        NativeLoader.init(new SystemDelegate());
+        initializeNativeLoaderWithSystemDelegate();
       } catch (IllegalAccessException e) {
         // Failed to initialize SoLoader
-        NativeLoader.init(new SystemDelegate());
+        initializeNativeLoaderWithSystemDelegate();
       } catch (InvocationTargetException e) {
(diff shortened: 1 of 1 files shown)


wizh:
Thanks for the PR!

alsutton:
No problem. If you can share details from intern phab about what the internal problems are I'd be happy to do a cleanup commit.

Issue #2463, opened 2020-02-08
color attribbute in fresco progress


saed220:
I use a vectodrwable progress that wrap in another drwable and for color of progress i use 
`?attr/my_attr`
when I use this progress in other views every things work but when I use it in fressco the atttr color 
disapear.
my vector drawble :
`<path
      android:pathData="M23.2,43.5c-11.9,0 -21.7,-9.7 -21.7,-21.7c0,-4.1 1.2,-8.1 3.4,-11.6c0.4,-0.7 1.4,-0.9 2.1,-0.5c0.7,0.4 0.9,1.4 0.5,2.1c-1.9,3 -2.9,6.4 -2.9,10c0,10.3 8.4,18.7 18.7,18.7c0.8,0 1.5,0.7 1.5,1.5C24.7,42.8 24,43.5 23.2,43.5z"
      android:fillColor="?attr/my_attr"/>`

saed220:
I also get error reraurse not found exception when I use drawable that contain attribbute color .
I think problem comes from this part of code  "getDrawable() "  in
 GenericDraweeHierarchy.java file of fresco.
I sugesct replace "getDrawable()" 
with "ContextCompat.getDrawable()"
but I dont know how could I change it ?

oprisnik:
Hey! Yeah, right now we do not use ContextCompat unfortunately. There's also https://github.com/facebook/fresco/issues/1463 - A pull request to migrate to ContextCompat would be appreciated :)

esafirm:
@saed220 is there any specific way to reproduce your issue?

PR #2583, opened 2021-02-07
Use ContextCompat to get drawable styled by context theme and update sample

realjustinwu:

## Motivation (required)

Closes(#2463 )

## Test Plan (required)

Open samples.showcase application, select "Attribute Color" item in drawer, the rectangle and a line of SVG file are use attribute color.


realjustinwu:
"context deadline exceeded", I have no permission to rerun ci job.

oprisnik (changes requested):
Please see my comment on the PR about having a new builder class.

oprisnik:
Hi! Thank you very much for the pull request. This is quite a big breaking API change (from Resources -> Context), so everyone using Fresco would have to refactor their code.
We're working on new image components right now and won't release a major new Drawee release. However, for your PR you can create a separate Drawee hierarchy builder with Context (based on what you're changing) and then that one can be used to create Context-based Drawables. This way, the API doesn't break and you can still use the Context version when required.

PR #2606, opened 2021-05-21
Migrate from JCenter to Maven Central publishing workflow

passy:
# Summary

Not quite ready yet:

- [x] Figure out how to generate javadoc. Currently failing because the default config pulls in *all* files, including non-Java ones, which the parser chokes on.
- [x] Only added `fbcore` to the list of artifacts to publish. Practically only requires adding the `apply plugin` line to the bottom of all other artifacts to publish.
- [x] Verify that the test "artifacts" that I removed weren't used anywhere.
- [x] Fix unit tests. (@oprisnik)
- [ ] Write docs.

# Test Plan:

```
./gradlew ./gradlew publishToMavenLocal -PRELEASE_SIGNING_ENABLED=false
```

Inspected local ~/.m2 and it looks good.

Did a `./gradlew publish` and checked the staging repo (dropped it afterwards - don't worry!)

Screenshot 2021-05-21 at 19 03 20

---

With new publish configs, published to my local repository:

```
ls -R ~/.m2/repository/com/facebook/fresco/ | pbcopy
```

https://gist.github.com/passy/c08a9438956dd9bca0aedb74666b955f


oprisnik:
This is amazing. Thanks a bunch :)

build.gradle
```diff
@@ -46,30 +45,37 @@ subprojects {
...
-                }
-            }
-
-            task copyDeps(dependsOn: setJarList, type: Copy) {
-                from jarList
-                into 'build/external'
-            }
-        }
-    }
+// TODO: Figure out why this breaks (and what it does).
```
passy, line 61: @oprisnik I'm not quite sure why this one breaks but also what it's there to do in the first place. 😅

Given that it's only run on cleanup, I suppose it's not super critical but would still be good to understand what the intent is.

It's failing now with this:

```
FAILURE: Build failed with an exception.

* Where:
Build file '/Users/realpassy/Projects/java/fresco/build.gradle' line: 65

* What went wrong:
A problem occurred configuring project ':animated-base'.
> Could not create task ':animated-base:uploadArchives'.
   > Task with name 'publishAllPublicationsToMavenCentralRepository' not found in project ':animated-base'.
```

oprisnik: I don't think we need this any more. Seems to be legacy code for Eclipse support :)

oprisnik (approved):

Issue #2605, opened 2021-05-17
Gif cache doesn't seem to be working.


WhileCrow:
### Description

Every time I load the same gif, all the frame caches for the gif are re-generated in memory. Is this a bug?Or GIF caching now not supported?

### Reproduction
```
class TestWebpActivity : AppCompatActivity() {
    private var mRecyclerView: RecyclerView? = null
    private val mTvNoResult: TextView? = null
    private var mAdapter: WebpAdapter? = null
    private val mDatas = ArrayList<String>()
    private var mContext: Context? = null


    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContentView(R.layout.activity_test_webplist)
        mRecyclerView = findViewById(R.id.recy)
        mContext = this
        testData()
        bindDatas()
    }

    private fun testData() {
        mDatas.add("https://upload.wikimedia.org/wikipedia/commons/thumb/2/2c/Rotating_earth_%28large%29.gif/200px-Rotating_earth_%28large%29.gif")
    }

    fun bindDatas() {
        mAdapter = WebpAdapter()
        mRecyclerView!!.layoutManager = LinearLayoutManager(this)
        mRecyclerView!!.setHasFixedSize(true)
        mRecyclerView!!.adapter = mAdapter
    }

    private inner class WebpAdapter : RecyclerView.Adapter<VH>() {
        private val items: MutableList<String> = ArrayList()
        override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): VH {
            return VH(LayoutInflater.from(mContext).inflate(R.layout.item_test_webplist, parent, false))
        }

        override fun getItemCount(): Int {
            return items.size
        }

        override fun onBindViewHolder(holder: VH, position: Int) {
            val url = items[position]
            holder.tv.text = url

            val width = 100
            val height = 100
            val request = ImageRequestBuilder.newBuilderWithSource(Uri.parse(url))
                    .setResizeOptions(ResizeOptions(width, height))
                    .build()
            val controller = Fresco.newDraweeControllerBuilder()
                    .setOldController(holder.img.getController())
                    .setImageRequest(request)
                    .setAutoPlayAnimations(true)
                    .build();
            holder.img.controller = controller



            if (url.contains(".gif")) {
                Thread {
                    while (true) {
                        Thread.sleep(1000)
                        Handler(mainLooper).post {
                            val height = 100
[...]

PR #2612, opened 2021-05-27
change imageId in AnimatedImageResult from the class hash to decodeHa…

WhileCrow:
I create a issue a few days ago, #2605, I'm kind of dying to know if this is a bug? And if this is, is my way of fixing it appropriate?

Test Plan
Load the same GIF multiple times, and :
1. Look at the Android profiler to see if the memory keeps increasingl
2. Debug if the count of entries in the cache pool keeps increasing.
Like the image in the #2605 shows.

### DIFF
--- a/animated-base/src/main/java/com/facebook/imagepipeline/animated/factory/AnimatedImageFactoryImpl.java
+++ b/animated-base/src/main/java/com/facebook/imagepipeline/animated/factory/AnimatedImageFactoryImpl.java
@@ -90,7 +90,7 @@ public class AnimatedImageFactoryImpl implements AnimatedImageFactory {
             sGifAnimatedImageDecoder.decodeFromNativeMemory(
                 input.getNativePtr(), input.size(), options);
       }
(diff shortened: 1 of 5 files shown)


WhileCrow (commented):
no review change

animated-base/src/main/java/com/facebook/imagepipeline/animated/factory/AnimatedImageFactoryImpl.java
```diff
@@ -90,7 +90,7 @@ public CloseableImage decodeGif(
             sGifAnimatedImageDecoder.decodeFromNativeMemory(
                 input.getNativePtr(), input.size(), options);
       }
-      return getCloseableImage(options, gifImage, bitmapConfig);
+      return getCloseableImage(bytesRef.getValueHash(), options, gifImage, bitmapConfig);
```
WhileCrow, line 93: maybe I should replace it with cache key

PR #2620, opened 2021-07-14
duplicate value for resource

MeandNi:
## Motivation (required)

duplicate value for 'attr/actualImageScaleType' in line 87 and 190.

- May cause runtime errors

> AAPT: error: duplicate value for resource 'attr/actualImageScaleType' with config ''.

## Test Plan (required)

Looking for the necessary path.


oprisnik:
Thanks, looks good. I'll import for internal testing.

oprisnik (approved):

PR #2663, opened 2022-03-18
fix load gallery video content uri failed

Huangjiaqi0421:
if not set android:requestLegacyExternalStorage="true" in Androidmanifest.xml,
there are no permission to access video path on Android Q.
we need use createThumbnailFromContentProvider() to get video thumbnail


steelrooter:
Stale PR. Closing.

PR #2760, opened 2024-02-25
feat: new downsample mode

sunnylqm:
## Motivation (required)

Introduce a new method, `setDownsampleMode(DownsampleMode)`, to allow downstream libraries to disable downsampling but not break other existing use cases. Fixes https://github.com/facebook/fresco/issues/2397

## Test Plan (required)

Test a large remote image like https://github.com/facebook/fresco/issues/2397#issuecomment-538223936
With default config or set `downsampleMode` to `AUTO/ALWAYS`, it should be downsampled/blurred like how it behaves today.
Change it to `NEVER` it should NOT be downsampled.


imagepipeline/src/main/java/com/facebook/imagepipeline/core/ProducerFactory.java
```diff
@@ -120,7 +120,7 @@ public ProducerFactory(
       ByteArrayPool byteArrayPool,
       ImageDecoder imageDecoder,
       ProgressiveJpegConfig progressiveJpegConfig,
-      boolean downsampleEnabled,
+      DownsampleMode downsampleMode,
```
yungsters, line 123: This is a breaking change (e.g. for callers or subclasses of `ProducerFactory`).

I think we need to add a constructor overload that continues accepting a `boolean isDownsampleEnabled` in this argument position.

yungsters (changes requested):

imagepipeline/src/main/java/com/facebook/imagepipeline/core/ProducerFactory.java
sunnylqm: added

sunnylqm:
@yungsters any more concerns?

yungsters:
I'm currently waiting for a colleague to take a look, but they're currently on vacation.

No concerns from me, but I'll report back as soon as we have progress.

yungsters:
@sunnylqm, this has been merged.

Thanks again for advocating for this improvement and for contributing this pull request that takes into consideration backward compatibility!

PR #2770, opened 2024-04-26
Upgrade androidx.core:core to 1.13.0

hick209:
Differential Revision: D56474755


oprisnik:
I don't think this one should be required any more @hick209 right?

hick209:
That's right @oprisnik, someone else already landed that change. Thanks for cleaning up this PR

PR #2780, opened 2024-07-29
Fix ndkVersion via Gradle scripts

steelrooter:
Differential Revision: D60339954


steelrooter:
Merged in 78fe67e2d3d01d699547eb3e250c2e690cad4a25

PR #2787, opened 2024-09-09
Add support for overriding downsample mode per image request

Abbondanzo:
## Summary

Fresco currently supports global configuration of downsampling modes via the `downsampleMode` setting in `ImagePipelineConfig`. However, it is not currently possible to support per-image downsample settings. Out of the box, the vast majority of images *should* be downsampled to avoid drawing images which are too large to be displayed on a canvas. Android will [throw a runtime exception](https://web.archive.org/web/20191017003524/https://chromium.googlesource.com/android_tools/+/refs/heads/master/sdk/sources/android-25/android/view/DisplayListCanvas.java#40) in such cases. Yet, there are some usages which warrant outright disabling of downsampling as detailed in great lengths in issues like [this one](https://github.com/facebook/react-native/issues/21301) and [this one](https://github.com/facebook/fresco/issues/2397). 

This change introduces a new mechanism for overriding the globally-configured `downsampleMode` per image request by using a new `downsampleOverride` setting. If the override value gets set to a non-null `DownsampleMode`, the `DecodeProducer` will use that override instead. If no override is set, the downsample mode configured from the pipeline is chosen.

## Examples

Here, you can see the subtle differences between two of the downsampling modes. These are applied via the new `downsampleOverride` setting. I have a follow-up PR that adds this to the showcase application.

|`NEVER`|`ALWAYS`|
|-|-|
|image|image|

Differential Revision: D62393210

### DIFF
--- a/imagepipeline/src/main/java/com/facebook/imagepipeline/core/DownsampleMode.kt
+++ b/imagepipeline/src/main/java/com/facebook/imagepipeline/core/DownsampleMode.kt
@@ -1,0 +1,14 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package com.facebook.imagepipeline.core
+
+enum class DownsampleMode {
+  ALWAYS,
+  AUTO,
+  NEVER
+}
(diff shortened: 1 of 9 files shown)


Abbondanzo:
Closed by 9c577385312f665b889d2957b43b77a84bd4e434

PR #2790, opened 2024-09-16
Minor fixes in image request, decoded images string builder

Abbondanzo:
Differential Revision: D62786986


Abbondanzo:
Landed in 95cb57e57d47e7a4840f5bb44f77f39b39ee6320

